PR #54, opened 2016-05-23
First pass e2e tests

raeeschachar:


raeeschachar:
@benpatterson @Muddasser 
Review?

requirements/base.txt
```diff
@@ -1,8 +1,255 @@
-selenium==2.39.0
```
benpatterson: I'm guessing this is going here because we'll be cleaning this up in a separate ticket?

benpatterson:
Hey @raeeschachar Just getting started on this, but I wanted to say that I don't think we should have a `stage/` directory. Ideally, we could re-use these tests in other environments (like sandboxes) depending on the configuration settings in `config.ini` (or elsewhere).

Could you make those changes? When you do, please don't squash commits until we are all set.

jenkins/test_edxapp.sh
```diff
@@ -24,12 +24,17 @@ set -x
 git clean -xfd
 
 # Create the virtualenv and install requirements
-mkdir -p venv
-virtualenv venv
-. venv/bin/activate
+#mkdir -p venv
```
benpatterson: I understand why you would not want to run these each time, but we'll need them somewhere so we can automate the test runs (even if it's push-button).

stage/pages/lms/course_page_lms.py
```diff
@@ -0,0 +1,9 @@
+from edxapp_pages.lms.course_info import CourseInfoPage
```
benpatterson: Makes sense, but should this be added to edx-platform (sooner or later)? It seems like a non-intrusive change.

stage/pages/lms/dashboard_lms.py
```diff
@@ -0,0 +1,15 @@
+from edxapp_pages.lms.dashboard import DashboardPage
+
+class DashboardPageExtended(DashboardPage):
+
+    def select_course(self):
+        """
+        Selects the course we want to perform tests on
+        Returns:
```
benpatterson: nit: delete `Returns:`

stage/pages/lms/dashboard_lms.py
```diff
@@ -0,0 +1,15 @@
...
+
+        """
+        course_names = self.q(css='.course-title a')
+        for vals in course_names:
+            if 'Manual Smoke Test Course 1 - Auto' in vals.text:
```
benpatterson: I think you should make this text a variable that you pass into the method. I also think you should raise some exception if the title is not found.

Something like this: 
```python
def select_course_by_title(self, course_title):
  course_names = self.q(css='.course-title a')
  for vals in course_names:
    if course_title in vals.text:
      vals.click()
  raise BrokenPromise("Course title not found") ## <-- no idea if that syntax is correct, but using it as an example
```

stage/tests/lms/test_dashboard.py
```diff
@@ -0,0 +1,26 @@
+from bok_choy.web_app_test import WebAppTest
+from ...pages.lms.login_lms import LmsLogin
+from ...pages.lms.dashboard_lms import DashboardPageExtended
+from ...pages.lms.course_page_lms import CourseInfoPageExtended
+from edxapp_pages.lms.courseware import CoursewarePage
+import os
```
benpatterson: nit: `import` statements should be on the top of the document, with `from x import y` below that.

stage/tests/lms/test_dashboard.py
```diff
@@ -0,0 +1,26 @@
+from bok_choy.web_app_test import WebAppTest
+from ...pages.lms.login_lms import LmsLogin
```
benpatterson: you should use fully-qualified import statements. See TE-1194

stage/tests/lms/test_dashboard.py
```diff
@@ -0,0 +1,26 @@
...
+from ...pages.lms.dashboard_lms import DashboardPageExtended
+from ...pages.lms.course_page_lms import CourseInfoPageExtended
+from edxapp_pages.lms.courseware import CoursewarePage
+import os
+
+class DashboardTest(WebAppTest):
+
+    DEMO_COURSE_USER = os.environ.get('USER_LOGIN_EMAIL')
```
benpatterson: This is good for now but we may need to put it into a common place, like a base class. No change needed for now; let's keep track of our vars though and plan to move them later.

stage/tests/lms/test_dashboard.py
```diff
@@ -0,0 +1,26 @@
...
+        super(DashboardTest, self).setUp()
+        self.login_page = LmsLogin(self.browser)
+        self.login_page.visit()
+        self.login_page.login(self.DEMO_COURSE_USER, self.DEMO_COURSE_PASSWORD)
+        self.dashboard_page = DashboardPageExtended(self.browser)
+
+    def test_resume_course(self):
+        self.dashboard_page.select_course()
+        course_page = CourseInfoPageExtended(self.browser, 'course-v1:ArbiRaees+AR-1000+fall')
+        course_page.is_browser_on_page()
```
benpatterson: I would refactor this... Below is a description:

* put the `course_page = ...` first
* instead of `course_page.is_browser_on_page()` use `course_page.wait_for_page()`. That has an implicit wait for `is_browser_on_page` to be true.

stage/tests/lms/test_dashboard.py
```diff
@@ -0,0 +1,26 @@
...
+        course_page.click_resume_button()
+        courseware_page = CoursewarePage(self.browser, 'course-v1:ArbiRaees+AR-1000+fall')
+        courseware_page.is_browser_on_page()
```
benpatterson: Same patterns here.... use the `wait_for_page`.

stage/tests/lms/test_demo.py
```diff
@@ -0,0 +1,25 @@
...
+from ...pages.lms.login_lms import LmsLogin
+from ...pages.lms.dashboard_lms import DashboardPageExtended
+
+
+class DemoTest(WebAppTest):
+    """
+    Sample Test for logging in and navigating to Courseware page
+    """
+
+    DEMO_COURSE_USER = os.environ.get('USER_LOGIN_EMAIL')
```
benpatterson: Ah I see we're already repeating this pattern. We should keep it somewhere else. How about a base class that sets up vars? In spite of what I said earlier, since we're already repeating a pattern, we should fix it up in this PR, not later.

stage/tests/studio/test_pages.py
```diff
@@ -4,21 +4,22 @@
...
 from edxapp_pages.studio.login import LoginPage
-from edxapp_pages.studio.manage_users import CourseTeamPage
+#from edxapp_pages.studio.manage_users import CourseTeamPage
 from edxapp_pages.studio.overview import CourseOutlinePage
 from edxapp_pages.studio.settings import SettingsPage
 from edxapp_pages.studio.settings_advanced import AdvancedSettingsPage
 from edxapp_pages.studio.settings_graders import GradingPage
 from edxapp_pages.studio.signup import SignupPage
 from edxapp_pages.studio.textbooks import TextbooksPage
+import os
```
benpatterson: nit: this should be above the `from ...` statements.

requirements/base.txt
raeeschachar: Yes this was the idea, as we have to change the requirements.

jenkins/test_edxapp.sh
raeeschachar: Done

stage/pages/lms/dashboard_lms.py
raeeschachar: Done.

stage/pages/lms/dashboard_lms.py
raeeschachar: Nice idea. This syntax works too. 
Done.

stage/tests/lms/test_dashboard.py
raeeschachar: Done.

stage/tests/lms/test_dashboard.py
raeeschachar: Done.

stage/tests/lms/test_dashboard.py
raeeschachar: Acknowledged.

stage/tests/lms/test_dashboard.py
raeeschachar: Done.

stage/tests/lms/test_dashboard.py
raeeschachar: Done. Will keep in mind next time, this is great learning thanks :)

stage/tests/studio/test_pages.py
raeeschachar: Done. 
This test has to be updated later on. This was the test written by Jay when he made this repo.

stage/tests/lms/test_demo.py
raeeschachar: I had a discussion with @Muddasser and it seems like he has already tackled this.

stage/pages/lms/course_page_lms.py
raeeschachar: Yes it makes sense to add it in edx-platform, should we add another ticket for this?

raeeschachar:
@benpatterson all comments addressed, one more look please?

pavelib/paver_tests/utils.py
```diff
@@ -0,0 +1,60 @@
...
+        super(PaverTestCase, self).setUp()
+
+        # Show full length diffs upon test failure
+        self.maxDiff = None  # pylint: disable=invalid-name
+
+        # Create a mock Paver environment
+        tasks.environment = MockEnvironment()
+
+        # Don't run pre-reqs
+        os.environ['NO_PREREQ_INSTALL'] = 'true'
```
benpatterson: Is this necessary in this environment? I don't think we're using prereqs in this repo are we?

pavelib/paver_utils.py
```diff
@@ -0,0 +1,74 @@
+from path import Path as path
+
+from paver_consts import LOG_DIR, TEST_DIR, REPORT_DIR, PAVER_TEST_DIR, PAVER_TEST_REPORT_DIR
+
+
+class NoseCommand(object):
+
+    def __init__(self):
```
benpatterson: You can just not-include this method, can't you?

pavelib/paver_tests/utils.py
benpatterson: Sorry..to be more precise, we're not using this environment variable...

regression/pages/lms/course_page_lms.py
```diff
@@ -0,0 +1,21 @@
...
+    where we add methods that are different or not used in CourseInfoPage
+    """
+    @property
+    def url(self):
+        """
+        Construct a URL to the page within the course.
+        """
+        return BASE_URL + "/courses/" + self.course_id + "/" + self.url_path
+
+    def click_resume_button(self):
```
benpatterson: Similar, I wonder if this could use used in the edx-platform version of this class. It wouldn't interfere with anything, would it?

regression/pages/lms/login_lms.py
```diff
@@ -0,0 +1,38 @@
...
+
+class LmsLogin(LoginPage):
+    """
+    This class is an extended class of LoginPage,
+    where we add methods that are different or not used in LoginPage
+    """
+    url = BASE_URL + '/login'
+
+    def is_browser_on_page(self):
+        return self.q(css='.action.action-primary.action-update.js-login.login-button').present
```
benpatterson: Should this be `.visible` instead? That'll be a stronger assertion if it applies.

regression/pages/lms/login_lms.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def provide_info(self, email, password):
+        """
+        Fill in login info
+        'Username' and 'Password' are the user's credentials
+        """
+        EmptyPromise(self.q(css='input#login-email').is_present, "Click ready").fulfill()
```
benpatterson: It's a  better pattern to use `wait_for` instead of `EmptyPromise`. (It's clearer to the reader.)

benpatterson: To avoid repeating `input#login-email`, you should set that in a variable at the top of the method, then reference the variable. Same goes for `login-password`.

regression/pages/__init__.py
```diff
@@ -0,0 +1,4 @@
+import os
+
+AUTH_USER_NAME = os.environ.get('BASIC_AUTH_USER', 'not_set')
```
benpatterson: nit: could you call the global variables `BASIC_AUTH_USERNAME` AND `BASIC_AUTH_PASSWORD`?

regression/tests/helpers.py
```diff
@@ -0,0 +1,47 @@
...
+    for page in pages:
+        print "Visiting: {}".format(page)
+        page.visit()
+
+
+class LoginHelper():
+    """
+    Helper class to login to the studio.
+    """
+    def __init__(self):
```
benpatterson: I don't think you need the `__init__` if it's not setting anything.

regression/tests/helpers.py
```diff
@@ -0,0 +1,47 @@
...
+def visit_all(pages):
+    """
+    Visit each page object in `pages` (an iterable).
+    """
+    for page in pages:
+        print "Visiting: {}".format(page)
+        page.visit()
+
+
+class LoginHelper():
```
benpatterson: This should really be `LoginHelper(object)`. We can get to that when we look at pylint.

regression/tests/lms/test_dashboard.py
```diff
@@ -0,0 +1,28 @@
...
+        super(DashboardTest, self).setUp()
+        self.login_page = LmsLogin(self.browser)
+        LoginHelper.login(self.login_page)
+        self.dashboard_page = DashboardPageExtended(self.browser)
+
+    def test_resume_course(self):
+        """
+        Verifies that we can successfully resume the course
+        """
+        course_page = CourseInfoPageExtended(self.browser, 'course-v1:ArbiRaees+AR-1000+fall')
```
benpatterson: You should set the coursename as a var to avoid repeating it.

regression/tests/lms/test_demo.py
```diff
@@ -0,0 +1,28 @@
...
+        super(DemoTest, self).setUp()
+        self.login_page = LmsLogin(self.browser)
+        self.dashboard_ext = DashboardPageExtended
+
+    def test_login(self):
+        """
+        Log in as a student
+        """
+        self.login_page.visit()
+        self.login_page.login(self.DEMO_COURSE_USER, self.DEMO_COURSE_PASSWORD)
```
benpatterson: You should have something after this, to assert that the login was successful.

benpatterson:
@raeeschachar I'm done with my second pass...thanks for the update.

raeeschachar:
Comments addressed. Good to go?

regression/tests/lms/test_demo.py
```diff
@@ -26,3 +26,5 @@ def test_login(self):
         """
         self.login_page.visit()
         self.login_page.login(self.DEMO_COURSE_USER, self.DEMO_COURSE_PASSWORD)
+        self.assertEquals(self.login_page.q(css='.wrapper-header-courses .header-courses').text[0].lower(),
```
benpatterson: A couple tweaks to this (and you should scan to see if you have similar patterns that I may have missed in previous code reviews)...
* replace `self.assertEquals` with `self.assertEqual`. The "Equals" with the 's' at the end is [deprecated](https://docs.python.org/2.7/library/unittest.html#deprecated-aliases).
* Set the 'Log-in successful' explicitly as the `msg` to make it clearer. Also, the message should reflect what the test should tell you if it fails, not if it's successful. So, something like this: `self.assertEqual(self.login_page.q...etc, 'my courses', msg='User not logged in as expected.')`

raeeschachar: @benpatterson I been thinking, should we remove this file? It was intended for the testing purposes.

benpatterson: Hmm. I imagined we were going to build it out over time. So I say leave it.

raeeschachar:
@benpatterson Done.

benpatterson:
LGTM 👍

raeeschachar:
@benpatterson Comments addressed after merging Pull 57
Can you go through it one more time? Good to go?

raeeschachar:
@benpatterson 
Fab file deleted and verified.

Muddasser:
👍

PR #55, opened 2016-05-23
Converted to paver from fab

Muddasser:


Muddasser:
@raeeschachar @benpatterson Please review.

pavelib/paver_consts.py
```diff
@@ -0,0 +1,11 @@
...
+
+# Paths for different directories
+LOG_DIR = path('./logs').abspath()
+TEST_DIR = path('./stage/tests/').abspath()
+REPORT_DIR = path('./reports').abspath()
+SCREENSHOT_DIR = path('./screenshots').abspath()
+BASELINE_DIR = path('./certs/screenshots/baseline').abspath()
+
+# XML reports names
+SMOKE_TEST_REPORT = '/smoke_tests_results.xml'
```
benpatterson: nit: don't use the directory indicator with this one (the forward slash).

benpatterson:
This looks like it has some of Raees's changes as well. Could you split it out? One way to do that is to make sure you have Raees's commit, and then make the PR against Raees's branch instead of master.

I gave it a quick look but would need to give it more in-depth look. So far it looks straightforward.

pavelib/paver_consts.py
Muddasser: Done.

Muddasser:
@benpatterson

PR #56, opened 2016-05-25
Integrated paver

Muddasser:


Muddasser:
@benpatterson @raeeschachar Please review.

raeeschachar:
@jzoldak Can you take a look too?

pavement.py
```diff
@@ -0,0 +1,54 @@
...
+    for env_var in env_vars:
+        try:
+            os.environ[env_var]
+        except:
+            raise BuildFailure("Please set the environment variable :" + env_var)
+
+
+@task
+def set_screenshots_path():
+    os.environ['NEEDLE_OUTPUT_DIR'] = SCREENSHOT_DIR
```
benpatterson: I think you can use `os.environ.get` instead, so it has a default that can be overridden if the end-user wants.

pavement.py
```diff
@@ -0,0 +1,54 @@
...
+    LOG_DIR.makedirs_p()
+
+
+@task
+def create_reports_directory():
+    REPORT_DIR.makedirs_p()
+
+
+@task
+@needs('check_env_vars', 'set_screenshots_path', 'create_log_directory', 'create_reports_directory')
```
benpatterson: I think you should have a task that sets all the paths. That task can call all of these.

pavement.py
```diff
@@ -0,0 +1,54 @@
...
+@consume_args
+def smoke_test(args):
```
benpatterson: Naming: I'd rather this be called something else, like `e2e_test`. When we have enough tests to warrant the difference between 'smoke' and 'regression', we can use something like nose attribs or paver args. But, for now let's make the name a bit more generic.

regression/pages/lms/__init__.py
```diff
@@ -0,0 +1,7 @@
+import os
+
+username = os.environ.get('BASIC_AUTH_USER', 'not_set')
```
benpatterson: you could also make this `''`.

regression/pages/studio/__init__.py
```diff
@@ -0,0 +1,8 @@
+import os
+
+username = os.environ.get('BASIC_AUTH_USER', 'not_set')
```
benpatterson: I see this is being repeated. Is there a common place we should put these?

regression/tests/helpers.py
```diff
@@ -9,3 +11,37 @@ def visit_all(pages):
...
+    """
+    def __init__(self):
+        pass
+
+    @staticmethod
+    def get_login_email():
+        """
+        Return login email set in environment variable
+        """
+        login_email = os.environ.get('USER_LOGIN_EMAIL', 'not_set@nothing.com')
```
benpatterson: I also feel like this logic should be generalized. (Perhaps I shoudln't be commenting on the tests...just paver?)

benpatterson:
@Muddasser I think this is coming along well.

I'd like to see tests for the paver commands. There are examples in edx-platform. I'd look at [these tests|https://github.com/edx/edx-platform/blob/master/pavelib/paver_tests/test_safelint.py]. Notice that they are using a common paver test base class. You should use that for mocking out the shell environment.

We also need to set up pep8 and pylint for this repo. I think we should do that after the fabric->paver conversion is done.

pavement.py
Muddasser: Ok. Do you want me to raise exception here then?  Actually I am bit confused. What are we trying to achieve here? This question arose when I saw your comment bellow: ‘you could also make this ' ' . Or, put more directly, is there a way to make it fail if these are not set?’.

regression/tests/helpers.py
Muddasser: No, you can comment on it. Generalized in what sense?

Muddasser:
@benpatterson I have updated the pull request. Some of the previous commented suggestions are not clear to me. I have replied to comments. 

@raeeschachar FYI.

benpatterson:
👍

PR #60, opened 2016-06-13
Raees/first studio tests with paver

raeeschachar:
- Some fixes from the Lms side first pass 
- Corrections on adding page objects 
- First Pass of Studio side tests 
All with paver. So the idea is that we are ready to add tests using paver now (once documentation is updated too). These make sure both Studio and Lms side tests work with Paver.


regression/pages/studio/login_studio.py
```diff
@@ -0,0 +1,19 @@
...
+        """
+        Attempt to log in using `email` and `password`.
+        """
+
+        self.q(css='input#email').fill(email)
+        self.q(css='input#password').fill(password)
+        self.q(css='button#submit').first.click()
+
+        # Ensure that we make it to Studio home page
+        EmptyPromise(
```
benpatterson: instead of `EmptyPromise` you should use `wait_for_page()` on a home page object.

regression/pages/studio/studio_home.py
```diff
@@ -0,0 +1,23 @@
...
+        Selects the course we want to perform tests on
+        Returns:
+
+        """
+        course_names = self.q(css=course_title)
+
+        for vals in course_names:
+            if 'Manual Smoke Test Course 1 - Auto' in vals.text:
+                vals.click()
+                break
```
benpatterson: rather than a `break`, you should have a `wait_for` of some sort to be sure the course has been selected.

Also, I suggest an `else` clause here, in case the course that was requested is not found. Perhaps raise a BrokenPromise error.

regression/tests/studio/test_studio_login.py
```diff
@@ -0,0 +1,28 @@
...
+    DEMO_COURSE_USER = os.environ.get('USER_LOGIN_EMAIL')
+    DEMO_COURSE_PASSWORD = os.environ.get('USER_LOGIN_PASSWORD')
+
+    def setUp(self):
+        """
+        Initialize the page object
+        """
+        super(StudioUserLogin, self).setUp()
+        self.studio_login_page = StudioLogin(self.browser)
+        self.studio_login_page.url = BASE_URL + '/signin'
```
benpatterson: you want this to exist on the `StudioLogin` class. I'll point out where.

regression/pages/studio/login_studio.py
```diff
@@ -0,0 +1,19 @@
+from bok_choy.promise import EmptyPromise
+from edxapp_pages.studio.login import LoginPage
+
+
+class StudioLogin(LoginPage):
```
benpatterson: in here you should set the url. [Here](https://github.com/edx/edx-platform/blob/master/common/test/acceptance/pages/lms/dashboard.py#L22)'s an example from platform.

regression/tests/studio/test_studio_home.py
```diff
@@ -0,0 +1,49 @@
...
+    DEMO_COURSE_USER = os.environ.get('USER_LOGIN_EMAIL')
+    DEMO_COURSE_PASSWORD = os.environ.get('USER_LOGIN_PASSWORD')
+
+    def setUp(self):
+        """
+        Initialize the page object
+        """
+        super(StudioHomeTest, self).setUp()
+        self.studio_login_page = StudioLogin(self.browser)
+        self.studio_login_page.url = BASE_URL + '/signin'
```
benpatterson: Similar comment here. You should set the `url` value on the `StudioLogin` class.

regression/tests/studio/test_studio_home.py
```diff
@@ -0,0 +1,49 @@
...
+        """
+        Verifies that user can select a course and navigate to its course outline page
+        """
+        self.studio_home_page = DashboardPageExtended(self.browser)
+
+        self.course_info = {
+            'org': 'test_org',
+            'number': self.unique_id,
+            'run': 'test_run',
+            'display_name': 'Test Course' + XSS_INJECTION + self.unique_id
```
benpatterson: Is this the case on stage?

regression/tests/studio/test_studio_home.py
```diff
@@ -0,0 +1,49 @@
...
+
+        self.studio_course_outline = CourseOutlinePage(
+            self.browser,
+            self.course_info['org'],
+            self.course_info['number'],
+            self.course_info['run']
+        )
+        self.studio_home_page.visit()
+        self.studio_home_page.select_course('.course-title')
+        self.studio_course_outline.is_browser_on_page()
```
benpatterson: you should do `wait_for_page` instead of `is_browser_on_page`

benpatterson:
@raeeschachar I'm done with my first pass. You should also rebase so that you can run travis for pep8, at leaset.

regression/pages/studio/login_studio.py
raeeschachar: The correction here was we didn't need this function as it was present in edx-platform. Done.

regression/tests/studio/test_studio_login.py
raeeschachar: Done.

regression/pages/studio/login_studio.py
raeeschachar: Done.

regression/tests/studio/test_studio_home.py
raeeschachar: Done. Nice suggestion. Thanks

regression/tests/studio/test_studio_home.py
raeeschachar: Yes. The only issue was browser authentication which was handled already. So we can use this now.

raeeschachar: As we are using edx platform's page objects, CourseOutlinePage there expects these arguments CourseOutlinePage(CoursePage, CourseOutlineContainer). We might have to use it in the future for Sections and Subsections. 
Fixed it according to our course.

regression/pages/studio/studio_home.py
raeeschachar: My bad. Done.

raeeschachar:
@benpatterson Addressed comments. One more look please.

raeeschachar:
Rebase didn't work

raeeschachar:
@benpatterson @Muddasser 
Fixed. Please review.

raeeschachar:
@Muddasser is there a way to check these tests locally instead of creating a PR and checking it here?

raeeschachar:
@benpatterson @Muddasser 
Please review.

regression/pages/studio/studio_home.py
```diff
@@ -0,0 +1,32 @@
...
+from regression.pages.studio import BASE_URL
+
+
+class DashboardPageExtended(DashboardPage):
+    """
+    This class is an extended class of Studio Dashboard Page,
+    where we add methods that are different or not used in DashboardPage
+    """
+
+    url = BASE_URL + '/home/'
```
benpatterson: do you need the second slash on `/home/`?

regression/tests/lms/test_dashboard.py
```diff
@@ -24,9 +25,9 @@ def test_resume_course(self):
         """
         Verifies that we can successfully resume the course
         """
-        course_name = 'course-v1:ArbiRaees+AR-1000+fall'
-        course_page = CourseInfoPageExtended(self.browser, course_name)
-        courseware_page = CoursewarePage(self.browser, course_name)
+        course_id = 'course-v1:ArbiRaees+AR-1000+fall'
```
benpatterson: Since you've declared this as a variable in the StudioHome page, can you import it, or put it in a shared location?

regression/tests/studio/test_studio_home.py
```diff
@@ -0,0 +1,46 @@
...
+        Initialize the page object
+        """
+        super(StudioHomeTest, self).setUp()
+        self.studio_login_page = StudioLogin(self.browser)
+        self.studio_login_page.visit()
+        self.studio_login_page.login(self.DEMO_COURSE_USER,
+                                     self.DEMO_COURSE_PASSWORD)
+        self.studio_home_page = DashboardPageExtended(self.browser)
+
+        self.course_info = {
```
benpatterson: There is more repetition here, albeit as a dict instead. I'm wondering if some of this can already be done with the edx-platform utils.

benpatterson: There is more repetition here, albeit as a dict instead. I'm wondering if some of this can already be done with the edx-platform utils.

requirements/base.txt
```diff
@@ -7,3 +7,5 @@ bok-choy==0.5.2
 pep8==1.5.7
 pylint==1.5.5
 edx-lint==0.5.1
+edx-opaque-keys==0.2.1
+mock==1.0.1
```
benpatterson: I see this...is it being used somewhere?

regression/pages/studio/studio_home.py
raeeschachar: Nope. Removed.

regression/tests/lms/test_dashboard.py
raeeschachar: Makes sense. Done.

regression/tests/studio/test_studio_home.py
raeeschachar: @benpatterson 
Not sure if I understand this correctly, but for stage we cannot use auto auth that we are normally using on edx-platform. We had to do this to use CourseOutlinePage in overview.py from edx-platform. Did you mean something else?

requirements/base.txt
raeeschachar: @Muddasser worked on requirements. This might came here due to rebasing. @Muddasser are we using it somewhere?

raeeschachar:
@benpatterson @Muddasser 
Added some replies and addressed comments. Please check.

requirements/base.txt
Muddasser: @raeeschachar This is being used. Some of the tests written require bokchoy pages using these requirements.

benpatterson: If that is the case, then those requirements should be included in the setup.py of the bok-choy page objects, rather than in this repo. Make sense?

regression/tests/studio/test_studio_home.py
benpatterson: No, I was really just pointing out that we have the course name/url in several places in the code. We should instead put it into a variable so any future changes are only changed in one place.

benpatterson: I see you have done that in helpers.py, so you should just use that variable here...

In order to use that variable, you may need to parse the string or use some sort of course_info method to parse it for you.

benpatterson:
Looks like this needs a rebase. Assuming there are no significant changes after that, you can merge from my perspective.

👍

Muddasser:
👍 Please rebase before merging.

PR #63, opened 2016-07-01
Test view live from Studio Dashboard page

raeeschachar:


raeeschachar:
@benpatterson @Muddasser 
Please review. This one was a difficult one. First, on clicking view live when it went to LMS, it displayed browser authentication. To handle that I added logging in to LMS in setup before logging in to Studio. Secondly, View Live button wasn't displaying unless we hover on to that, I tried move_to_element and ActionChains from selenium but it didn't work, so we had to use execute_script to handle this.

regression/pages/studio/studio_home.py
```diff
@@ -39,3 +39,12 @@ def click_logout_button(self):
         self.wait_for_element_visibility(
             '.action-signout', 'Sign out button visibility')
         self.q(css='.action-signout').click()
+
+    def click_view_live_button(self):
+        """
+        Clicks view live button as per sent course_id
+        """
+        self.browser.execute_script(
+            "document.querySelectorAll('[data-course-key = \"course-v1:"
```
benpatterson: The course key should be extracted as a variable...

regression/tests/studio/test_studio_home.py
```diff
@@ -43,3 +46,44 @@ def test_studio_course_select(self):
...
+        LoginHelper.login(self.login_page)
+
+        self.studio_login_page = StudioLogin(self.browser)
+        self.studio_login_page.visit()
+        self.studio_login_page.login(self.DEMO_COURSE_USER,
+                                     self.DEMO_COURSE_PASSWORD)
+        self.studio_home_page = DashboardPageExtended(self.browser)
+
+        self.course_info = {
+            'org': 'ArbiRaees', 'number': 'AR-1000', 'run': 'fall'}
```
benpatterson: Same here....this should be a var so it's in fewer places in the code... I also think it makes sense to have it as an environment var

PR #64, opened 2016-07-01
Course update tests

Muddasser:


Muddasser:
@jzoldak @benpatterson @muhammad-ammar Please review.

regression/pages/studio/course_info_studio.py
```diff
@@ -0,0 +1,61 @@
...
+        click_css(self, '.button.new-button.new-update-button', 0, False)
+        self.wait_for_element_presence(
+            '.new-update-form', 'Update form has been opened')
+
+    def write_update_and_save(self, new_update):
+        """
+        Write new update and save it.
+        """
+        type_in_codemirror(self, 0, new_update)
+        click_css(self, '.save-button', 0, True)
```
benpatterson: Do we need a `wait_for` condition after this click, to ensure the page is done painting and is ready for the assertion or whatever else comes next?

regression/pages/studio/course_info_studio.py
```diff
@@ -0,0 +1,61 @@
...
+        self.wait_for_element_presence(
+            '.new-update-form', 'Update form has been opened')
+        self.write_update_and_save(course_update_edit_text)
+
+    def delete_course_update(self, index=0):
+        """
+        Delete a course update.
+        """
+        click_css(self, '#course-update-view .delete-button', index, False)
+        click_css(self, '.prompt.warning.has-actions .action-primary', 0, True)
```
benpatterson: I think a `wait_for` condition needs to follow this line as well.

regression/pages/studio/course_info_studio.py
```diff
@@ -0,0 +1,61 @@
...
+    def edit_course_handout(self, update_handout_text):
+        """
+        Edit course handout.
+        """
+        click_css(
+            self, '#course-handouts-view .edit-button', 0, False)
+        self.wait_for_element_visibility(
+            '.edit-handouts-form', 'Handout edit form visible.')
+        type_in_codemirror(self, 0, update_handout_text)
+        click_css(self, '.save-button', 0, True)
```
benpatterson: Another `wait_for` condition here?

regression/pages/studio/course_page_studio.py
```diff
@@ -0,0 +1,24 @@
...
+from regression.pages.studio.utils import get_course_key
+from regression.pages.studio import BASE_URL
+
+
+class CoursePageExtended(CoursePage):
+    """
+    Extended CoursePage.
+    """
+    # Needs to be overridden by child classes.
+    def is_browser_on_page(self):
```
benpatterson: In python 2.7, you can enforce this with an `@abstractmethod` annotation.

regression/pages/studio/login_studio.py
```diff
@@ -0,0 +1,30 @@
...
+from bok_choy.promise import EmptyPromise
+from edxapp_acceptance.pages.studio.login import LoginPage
+
+from regression.pages.studio import BASE_URL
+
+
+class StudioLogin(LoginPage):
+    """
+    This class is an extended class of LoginPage.
+    We are over-ridding methods that are different or adding new
```
benpatterson: nit:s/over-ridding/overriding

regression/pages/studio/login_studio.py
```diff
@@ -0,0 +1,30 @@
...
+    def login(self, email, password):
+        """
+        Attempt to log in using `email` and `password`.
+        """
+        self.q(css='input#email').fill(email)
+        self.q(css='input#password').fill(password)
+        self.q(css='button#submit').first.click()
+
+        # Ensure that we make it to another page
+        EmptyPromise(
```
benpatterson: Instead of empty promise, you should use `wait_for()`

regression/pages/studio/utils.py
```diff
@@ -0,0 +1,18 @@
+"""
+Utility functions for studio page objects.
+"""
+from opaque_keys.edx.locator import CourseLocator
+
+
+def get_course_key(course_info, module_store='split'):
+    """
+    Returns the course key based upon course info passed.
+    Course key depends upon the module store passed. The default
```
benpatterson: This docstring confuses me a bit. Is the course key different if it is split vs draft?

regression/tests/helpers.py
```diff
@@ -4,6 +4,19 @@
 import os
 
 
+def get_course_info():
+    """
+    Returns the course info of the course that we use for
+    the regression tests.
+    """
+    return {
```
benpatterson: I think this should be an environment var or something like that. We may decide to change which course we're using, and we should be able to do that without modifying code...

regression/tests/studio/test_course_update.py
```diff
@@ -0,0 +1,86 @@
...
+        # Write update and save it.
+        self.course_update_page.open_new_update_form()
+        self.course_update_page.write_update_and_save(course_update_text)
+        # Verify that the new update has been saved correctly and is visible.
+        course_update_content_selector = '#course-update-list li' \
+                                         ' .post-preview .update-contents'
+        self.assertEqual(
+            self.course_update_page.q(
+                css=course_update_content_selector)[0].text,
+            course_update_text
```
benpatterson: could you explain the assertion? What is it we want to see that's equal?

regression/tests/studio/test_course_update.py
```diff
@@ -0,0 +1,86 @@
...
+                self.course_update_page.q(
+                    css=course_update_content_selector)[0].text,
+                course_update_text
+            )
+
+    def test_edit_course_handout(self):
+        """
+        Test edit course handout
+        """
+        course_handout_content = 'New handout content:{}'.format(uuid4().hex)
```
benpatterson: I like that this has a unique ID. But could we delete it when the test finishes, anyway?

regression/pages/studio/course_info_studio.py
Muddasser: You are right. Just for clarification, click_css waits for the ajax calls and also ensures save notification is shown (if last argument is True) before returning. This was the reason I didn't use waits myself. But again you are right, we should be vigilant of any flaky conditions it may cause in the future. I have added waits to all clicks as asked in the review.

regression/pages/studio/utils.py
Muddasser: Yes, it would be different. For example, when we use 'split' eventual url in the page would be studio.stage.edx.org/course_info/course-v1:ArbiRaees+AR-1000+fall but  studio.stage.edx.org/course_info/ArbiRaees/AR-1000/fall when using 'draft'. studio.stage.edx.org/course_info/ArbiRaees/AR-1000/fall (The url got using 'draft') would cause 403 forbidden error.

regression/tests/studio/test_course_update.py
Muddasser: We are ensuring that the course update has been successful. `course_update_text` is the text we want to store as a new update. `self.course_update_page.q(css=course_update_content_selector)[0].text` gets the update text from the most recent update and we assert that it is equal to `course_update_text`.

regression/tests/helpers.py
Muddasser: I have changed it to get values from environment variables. Thought default is still set to the course used before.

Muddasser:
@benpatterson Please take a look.

regression/pages/studio/utils.py
```diff
@@ -0,0 +1,18 @@
+"""
+Utility functions for studio page objects.
+"""
+from opaque_keys.edx.locator import CourseLocator
+
+
+def get_course_key(course_info, module_store='split'):
```
Muddasser: @benpatterson I am thinking of getting module_store info from environment variable instead of passing it as an argument.  What are your suggestions?

regression/tests/helpers.py
```diff
@@ -4,16 +4,23 @@
...
     """
     Returns the course info of the course that we use for
     the regression tests.
     """
     return {
-        'org': 'ArbiRaees',
-        'number': 'AR-1000',
-        'run': 'fall',
-        'display_name': 'Manual Smoke Test Course 1 - Auto'
+        'org': os.environ.get(COURSE_ORG, 'ArbiRaees'),
```
benpatterson: I think that you should have it default to None, rather than the values you want. The end-result should be that the user sets the values before running the test. If you prefer, you can store these in a separate .ini file or something along those lines.

PR #68, opened 2016-07-29
Studio textbook page tests for e2e repo

raeeschachar:


raeeschachar:
@benpatterson @Muddasser 
This was the setup format I was discussing in the meeting yesterday (we can use the old one too tho). In setup,
`self.studio_home_page.q(css='.nav-item.nav-course-courseware').click()
self.studio_home_page.q(css='.nav-item.nav-course-courseware-textbooks').click()
self.textbook_page.wait_for_page()`
this makes sure we are manually clicking the drop down and clicking the textbooks page and navigating to that, so any bugs related to this won't be missed. 
Please review and let me know if we should add one test for all drop downs and navigating or we should structure the setup this way for the tests.

regression/tests/studio/test_studio_textbooks.py
```diff
@@ -0,0 +1,72 @@
...
+            self.course_info['number'],
+            self.course_info['run']
+        )
+
+        self.studio_home_page.visit()
+        self.studio_home_page.select_course(
+            'Manual Smoke Test Course 1 - Auto')
+        self.studio_course_outline.wait_for_page()
+
+        self.studio_home_page.q(css='.nav-item.nav-course-courseware').click()
```
benpatterson: I think you should only do this kind of navigation testing in one test for the thing you are testing, not all of them. It can turn into a lot of time waste if we use it too much; we should only use it for the time that it's important to test the navigation (i.e., only one test per feature, or something like that).

raeeschachar: Makes sense. I will fix it and send an update.

raeeschachar:
@benpatterson @Muddasser 
Fixed URL as discussed. Please review.

regression/pages/studio/login_studio.py
```diff
@@ -13,14 +14,23 @@ class StudioLogin(LoginPage):
...
+    def login(self, email, password, expect_success=True):
+        """
+        Attempt to log in using 'email' and 'password'.
+        """
+        self.fill_field('input#email', email)
+        self.fill_field('input#password', password)
         self.q(css='button#submit').first.click()
 
         # Ensure that we make it to another page
-        self.wait_for(lambda: 'signin' not in self.browser.current_url,
```
benpatterson: I made a similar comment on a different PR about `wait_for` being the preferred pattern instead of `EmptyPromise`.  I wonder if you need a rebase or a separate PR for login?

benpatterson:
This looks ilke code I just reviewed in #69. I'm wondering if there needs to be a rebase?

raeeschachar:
I did rebase but it seems like I rebased incorrectly. Yes it is the same code, actually this 68 is also in 69 after my rebase. I will close this one and we can continue on 69. Thanks

PR #71, opened 2016-08-11
Updated shell script for jenkins

Muddasser:


Muddasser:
@benpatterson @raeeschachar 
I have changed shell script for the Jenkins. Also there are some changes in pavement file to run the tests. I am merging this PR as is. If we find any problem, we can always come back to it latter.

Muddasser:
We are changing it for running the job on Jenkins.

PR #74, opened 2016-08-19
HTML Component tests

raeeschachar:


raeeschachar:
@benpatterson @Muddasser @muhammad-ammar 
Please review.

regression/pages/studio/unit_page.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+    def view_live_version(self):
+        """
+        Clicks View live version button on Units page
+        """
+        self.q(
+            css='.action-button'
+            '[title="Open the courseware in the LMS"]'
+        ).click()
+        self.browser.switch_to_window(self.browser.window_handles[-1])
```
benpatterson: I'm not sure why this is important (I'm sure it is though). Could you give some details in the docstring? (Something like "click this button, which opens up a new window".)

regression/tests/common/test_html_component.py
```diff
@@ -0,0 +1,115 @@
...
+    DEMO_COURSE_PASSWORD = os.environ.get('USER_LOGIN_PASSWORD')
+
+    def setUp(self):
+        """
+        Initialize the page object
+        """
+        super(StudioLmsHTMLTest, self).setUp()
+        # Login to Lms first to avoid authentication problems
+        self.login_page = LmsLogin(self.browser)
+        LoginHelper.login(self.login_page)
```
benpatterson: I forget; if the user is already logged in, will this fail? IIRC selenium will reset your cookies with each run, but I'm not positive on that at the moment.

regression/tests/common/test_html_component.py
```diff
@@ -0,0 +1,115 @@
...
+
+        self.unit_container_page = UnitPageExtended(
+            self.browser, None
+        )
+
+        self.studio_login_page = StudioLogin(self.browser)
+        self.studio_login_page.visit()
+        self.studio_login_page.login(self.DEMO_COURSE_USER,
```
benpatterson: Same question here

regression/tests/common/test_html_component.py
```diff
@@ -0,0 +1,115 @@
...
+            self.course_info['run'])
+
+        self.lms_courseware = CoursewarePage(self.browser, self.course_info)
+
+    def test_html_components(self):
+        """
+        Verifies that user can add HTML components on Studio and LMS
+        """
+        self.studio_course_outline.visit()
+        self.studio_course_outline.add_section_with_name('Test Section 1')
```
benpatterson: I wonder if it would be a good idea to append a uuid to this string to avoid any collisions between test runs.

benpatterson:
Done with my first pass. The uuid I think you should do to avoid some tech debt.

regression/tests/common/test_html_component.py
raeeschachar: It does I think. 
What this is doing is, on stage, when we have to use View Live (navigating from Studio to LMS), windows authentication box appears, which breaks the test. To avoid that we login to LMS first, then on Studio, perform our tasks and then if we click View Live (navigate from Studio to LMS, it directly goes to the page without asking for windows authentication. By windows authentication I mean stage server security.

raeeschachar:
Comments addressed. Please review. 
@benpatterson

Muddasser:
👍

benpatterson:
👍

PR #83, opened 2016-10-03
End to end script changes for Jenkin jobs  TE-1664

jzoldak:
@edx/testeng 
* Set the DISPLAY to Xvfb so that firefox can run
* Rename log dir envar to match the folder structure
 
Notes for @Muddasser
-
* This should get you back in action on [the jenkins job](https://build.testeng.edx.org/job/edx-e2e-tests)
* It is a private job in jenkins so you need to be logged in to see it
* Screenshots, browser logs, etc are captured on failure. To see them drill through on the "Build Artifacts" from a job, and look in the "log" folder
* A number of the tests are failing, I think this is due to the user that we have configured not being enrolled in the course under test. We can update the credentials - that will be a change in the edx-ops/testeng-secure repo. Let me know via email or chat what to use instead and I'll get that PR through.


jenkins/end_to_end_tests
```diff
@@ -1,25 +0,0 @@
-#!/usr/bin/env bash
```
jzoldak: reviewer note: this is NOT the script file that we use, but instead it is a file with no extension that got checked in accidentally.

pavelib/paver_consts.py
```diff
@@ -4,7 +4,7 @@
 
 # Constants for e2e tests
 E2E_TEST_REPORT = 'e2e_tests_results.xml'
-LOG_DIR = path(os.path.join('logs')).abspath()
+LOG_DIR = path(os.path.join('log')).abspath()
```
jzoldak: reviewer note: 
The jenkins job is set up to archive the "log" directory, not the "logs" directory. This was a typo.

jenkins/end_to_end_tests.sh
```diff
@@ -17,7 +17,14 @@ virtualenv venv
...
 
-# Install pages.
+# Install the page objects from the edx-platform repo.
+# Before doing so, we don't need optimizations for lxml,
+# so install it this way which doesn't bother compiling them.
+STATIC_DEPS=true CFLAGS="-O0"  pip install "lxml==3.4.4"
 paver install_pages
+
+# Set the display to the virtual frame buffer (Xvfb)
+export DISPLAY=:1
```
jzoldak: Reviewer note: This was why firefox was erroring instead of starting, it had nowhere to draw the display.
In the edx-platform repo jobs, this is handled by the [jenkins-common script](https://github.com/edx/edx-platform/blob/master/scripts/jenkins-common.sh#L3), which sources [a file that was put in the jenkins user's home directory](https://github.com/edx/configuration/blob/master/playbooks/roles/jenkins_worker/templates/jenkins_env.j2#L5) on the jenkins-worker during its initial configuration.

jenkins/end_to_end_tests.sh
```diff
@@ -17,7 +17,14 @@ virtualenv venv
 
 pip install -r requirements/base.txt
 
-# Install pages.
+# Install the page objects from the edx-platform repo.
+# Before doing so, we don't need optimizations for lxml,
+# so install it this way which doesn't bother compiling them.
+STATIC_DEPS=true CFLAGS="-O0"  pip install "lxml==3.4.4"
```
benpatterson: this is a nice pattern; you should tell other folks about it if they are waiting on installs.

jenkins/end_to_end_tests.sh
benpatterson: I wonder if we want to do anything here to improve bok-choy logging so it's clearer when this error occurs. I'm not sure if we can get enough information from selenium, but thought I'd ask.

jzoldak: @benpatterson 
**tldr:** I looked into it but called it too much feature creep from the current issue.

**longer:** I figured out the root cause with firefox logging, and started putting in the changes to make that an option but although it's pretty easy even that wasn't a gimme. https://github.com/edx/bok-choy/pull/177

BTW I forget what the firefox output was, but IIRC it was something along the lines of "Could not open the display" which was pretty much a flashing neon sign.

PR #88, opened 2016-10-13
Utils

kashifch:
@benpatterson 

I am breaking the PR in smaller chunks.
Here is the first installments
It basically consists of Utils
I have made following changes
Added unit tests to most of the methods
Brought all regex functions from across tests  in Utils and added unit tests for these
The date conversion function is now generic
I have removed the Email APIs I was using previously, now I am creating new users and reading emails using the python imap plus gmail aliases
The course ids and numbers are now extracted using opaque keys

Things which are not changed
I haven't yet changed the skip_clean_up, as I have commented earlier, I am looking into using the api for un-enrolling multiple users from course, if it works I will remove the function, otherwise we could look into an alternate.

I am still clicking the checkboxes/option buttons using javascript as I experience flakiness when I use the selenium click, if that is unacceptable I can look into other options

### DIFF
--- a/regression/pages/common/tests/test_utils.py
+++ b/regression/pages/common/tests/test_utils.py
@@ -1,0 +1,476 @@
+"""
+Unit tests for utils
+"""
+import unittest
+import textwrap
+
+from ddt import ddt, data, unpack
+
+from regression.pages.common.utils import (
+    get_target_url_from_text,
+    read_enrollment_codes_from_text,
+    extract_mmm_dd_yyyy_date_string_from_text,
+    convert_date_format,
+    extract_numerical_value_from_price_string
+)
+
+
+@ddt
+class TestUtils(unittest.TestCase):
+    """
+    Class for Utils unit tests
+    """
+
+    @data(
+        {
+            'partial_string': 'activate',
+            'text_chunk': """
+            Thank you for creating an account with our site.
+
+            There's just one more step before you can enroll in a course: you
+            need to activate your example organization account.
+            To activate your account, click the following link. If that
+            doesn't work, copy and paste the link into your browser's
+            address bar.
+
+            https://example-site.edu/activate/f290d7a1745a4839b8.
+            If you didn't create an account, you don't need to do anything;
+            you won't receive any more email from us. If you need assistance,
+            please do not reply to this email message. Check the help section
+            of the Organization website
+            """,
+            'result':
+                'https://example-site.edu/activate/f290d7a1745a4839b8'
+        },
+        {
+            'partial_string': 'password_reset_confirm',
+            'text_chunk': """
+            You're receiving this e-mail because you requested a password
+            reset for your user account at example.edu.
+
+            Please go to the following page and choose a new password:
+
+            https://example-site.edu/password_reset_confirm/336nr-4/
+
+
+            If you didn't request this change, you can disregard this email -
+            we have not yet reset your password.
+
+            Thanks for using our site!
+
+            The Example Organization Team
+            """,
+            'result':
+                'https://example-site.edu/password_reset_confirm/336nr-4/'
+        },
+        {
+            'partial_string': 'enrollment_code_csv',
+            'text_chunk': """
+            Order confirmation for: EXM-123
+
+            Dear User,
+
+            Thank you for purchasing access to Example Course. Please go to
+            https://example-site.edu/coupons/enrollment_code_csv/EXM-123/
+            to download a CSV file with the enrollment codes for this course.
+            Once you have the codes you can distribute them to your team.
+
+            To explore other courses,
+            please visit https://example-site.edu.
+
+            Thank you,
+            Example Organization Team
+
+            You received this message because you purchased enrollment codes
+            for Example course on https://example.edu.
+            If you have any questions, please visit
+            https://example-site.edu/contact.
+            """,
+            'result':
+                'https://example-site.edu/coupons/enrollment_code_csv/EXM-123/'
+        }
+    )
+    @unpack
+    def test_get_target_url_from_text_with_valid_data(
+            self,
+            partial_string,
+            text_chunk,
+            result
+    ):
+        """
+        Test that regex is correctly extracting the target url based on
+        partial string in different cases.
+        Args:
+            partial_string:
+            text_chunk:
+            result:
+        """
+        self.assertEqual(
+            get_target_url_from_text(partial_string, text_chunk),
+            result
+        )
+
+    @data(
+        {
+            'partial_string': 'invalid_string',
+            'text_chunk': """
+            Order confirmation for: EXM-123
+
+            Dear User,
+
+            Thank you for purchasing access to Example Course. Please go to
+            https://example-site.edu/coupons/enrollment_code_csv/EXM-123/
+            to download a CSV file with the enrollment codes for this course.
+            Once you have the codes you can distribute them to your team.
+
+            To explore other courses,
+            please visit https://example-site.edu.
+
+            Thank you,
+            Example Organization Team
+
+            You received this message because you purchased enrollment codes
+            for Example course on https://example.edu.
+            If you have any questions, please visit
+            https://example-site.edu/contact.
+            """
+        },
+        {
+            'partial_string': 'activate',
+            'text_chunk': """
+            Some text without URL
+            """
+        }
+    )
+    @unpack
+    def test_get_target_url_from_text_with_invalid_data(
+            self,
+            partial_string,
+            text_chunk
+    ):
+        """
+        Test that in case the regex search is not successful a
+        meaningful message is displayed
+        Args:
+            partial_string:
+            text_chunk:
+        """
+        self.assertEqual(
+            get_target_url_from_text(partial_string, text_chunk),
+            'Target URL not found in the text'
+        )
+
+    @data(
+        {
+            'source_data': textwrap.dedent("""\
+            Order Number:,EXAMPLE-116375
+
+            Enrollment code for professional seat in course
+            Code,Redemption URL
+            AWMHEUSMXYZDJXRN,https://coupons/?code=AWMHEUSMXYZDJXRN
+            PZBEQDSAVH26PPH4,https://coupons/?code=PZBEQDSAVH26PPH4
+            DNMV2T6PJAUVZUHC,https://coupons/?code=DNMV2T6PJAUVZUHC"""),
+            'result':
+            {
+                'AWMHEUSMXYZDJXRN': 'https://coupons/?code=AWMHEUSMXYZDJXRN',
+                'PZBEQDSAVH26PPH4': 'https://coupons/?code=PZBEQDSAVH26PPH4',
+                'DNMV2T6PJAUVZUHC': 'https://coupons/?code=DNMV2T6PJAUVZUHC'
+            },
+        },
+        {
+            'source_data': textwrap.dedent("""\
+            Order Number:,EXAMPLE-116375
+
+            Enrollment code for professional seat in course
+            Code,Redemption URL
+            AWMHEUSMXYZDJXRN,https://coupons/?code=AWMHEUSMXYZDJXRN
+            PZBEQDSAVH26PPH4,https://coupons/?code=PZBEQDSAVH26PPH4
+            DNMV2T6PJAUVZUHC,https://coupons/?code=DNMV2T6PJAUVZUHC"""),
+            'result':
+            {
+                'AWMHEUSMXYZDJXRN': 'https://coupons/?code=AWMHEUSMXYZDJXRN',
+                'PZBEQDSAVH26PPH4': 'https://coupons/?code=PZBEQDSAVH26PPH4',
+                'DNMV2T6PJAUVZUHC': 'https://coupons/?code=DNMV2T6PJAUVZUHC'
+            }
+        }
+    )
+    @unpack
+    def test_get_enrollment_codes_from_email_with_valid_data(
+            self,
+            source_data,
+            result
+    ):
+        """
+        Test that coupon codes are read successfully from email
+        Args:
+            source_data:
+            result:
+        """
+        self.assertDictEqual(
+            read_enrollment_codes_from_text(source_data),
+            result
+        )
+
+    @data(
+        """
+        Order Number:,MITPE-116375
+
+        Enrollment code for invalid course
+        """,
+        textwrap.dedent("""\
+        Order Number:,MITPE-116375
+
+        Enrollment code for invalid course
+        abc123,https://coupons/?code=abc
+        yth765,https://coupons/?code=thtg
+        """)
+    )
+    def test_get_enrollment_codes_from_email_with_invalid_data(
+            self,
+            source_data
+    ):
+        """
+        Test that a relevant error message is displayed when coupons are not
+        found
+        Args:
+            source_data:
+        """
+        self.assertEqual(
+            read_enrollment_codes_from_text(source_data),
+            'Coupons not found'
+        )
+
+    @data(
+        {
+            'original_string': 'Course starts: May 04, 2016',
+            'result': 'May 04, 2016'
+        },
+        {
+            'original_string': 'Discount valid until Oct 31, 2016',
+            'result': 'Oct 31, 2016'
+        }
+    )
+    @unpack
+    def test_extract_mmm_dd_yyyy_date_string_from_text_with_valid_data(
+            self,
+            original_string,
+            result
+    ):
+        """
+        Test that date of the format MMM dd, yyy can be extracted successfully
+        from teh target string
+        Args:
+            original_string:
+            result:
+        """
+        self.assertEqual(
+            extract_mmm_dd_yyyy_date_string_from_text(original_string),
+            result
+        )
+
+    @data('Course Start: 07-12-15', 'No date')
+    def test_extract_mmm_dd_yyyy_date_string_from_text_with_invalid_data(
+            self,
+            original_string
+    ):
+        """
+        Test in case the date is not found a relevant error message is raised
+        Args:
+            original_string:
+        """
+        self.assertEqual(
+            extract_mmm_dd_yyyy_date_string_from_text(original_string),
+            'Required date pattern not found in search string'
+        )
+
+    @data(
+        {
+            'original_date': 'Oct 31, 2016',
+            'original_format': '%b %d, %Y',
+            'required_format': '%Y-%m-%dT%H:%M:%S',
+            'result': '2016-10-31T00:00:00'
+        },
+        {
+            'original_date': '2016-10-31T00:00:00',
+            'original_format': '%Y-%m-%dT%H:%M:%S',
+            'required_format': '%b %d, %Y',
+            'result': 'Oct 31, 2016'
+        },
+        {
+            'original_date': '2013-1-25',
+            'original_format': '%Y-%m-%d',
+            'required_format': '%m/%d/%y',
+            'result': '01/25/13'
+        },
+        {
+            'original_date': '15/2/2010',
+            'original_format': '%d/%m/%Y',
+            'required_format': '%a %b %d %Y',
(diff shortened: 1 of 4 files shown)


Muddasser:

regression/pages/common/tests/test_consts.py
```diff
@@ -0,0 +1,311 @@
...
+
+ENROLLMENT_CODES_FROM_EMAIL = {
+    'valid_data': {
+        'source_data': """
+        Order Number:,MITPE-116375
+
+        Enrollment code for professional seat in Internet of Things: Roadmap
+        to a Connected World
+        Code,Redemption URL
+        AWMHEUSMXYZDJXRN,https://payments.mitprofessionalx-stage.mit.edu/
```
benpatterson: Are these real keys and values? Just a reminder that this is open source...

regression/pages/common/tests/test_utils.py
```diff
@@ -0,0 +1,254 @@
...
+        """
+        Test that dates can be changed from one format to other successfully
+        Also in case of any error a relevant error message is displayed
+        Args:
+            original_date:
+            original_format:
+            required_format:
+            result:
+
+        Returns:
```
benpatterson: You don't need the `Returns` info in the docstrings for unit tests

regression/pages/common/utils.py
```diff
@@ -0,0 +1,496 @@
...
+    pass
+
+
+class MailClient(object):
+    """
+    Connect to Gmail clinet using imap and read mails from Inbox
+    """
+
+    def __init__(self):
+        self.mail = imaplib.IMAP4_SSL('imap.gmail.com')
```
benpatterson: The client should be a variable you set somewhere, whether it's the top of the file or in the constants file. I'll re-read your arguments on why we're not using a test service.... I don't like relying on a third party for such a critical part of tests.

regression/pages/common/utils.py
```diff
@@ -0,0 +1,496 @@
...
+        for item in items:
+            coupon_codes.append(item['code'])
+        return coupon_codes
+
+
+def yesterday_date():
+    """
+    Get and return yesterday's date in dd-mmm-yyyy format
+    """
+    return (datetime.date.today() - datetime.timedelta(1)).strftime("%d-%b-%Y")
```
benpatterson: is timezone important? I'm assuming yes and wondering how that plays into this.

regression/pages/common/utils.py
```diff
@@ -0,0 +1,496 @@
...
+            if row:
+                new_row = row.split(',')
+                coupons.update({new_row[0]: new_row[1]})
+    if coupons:
+        return coupons
+    else:
+        return 'Coupons not found'
+
+
+def skip_cleanup_if_test_passed():
```
benpatterson: Cleanup should be idempotent. You should be able to run it whether the test passed or failed. Its job is to ensure a given end-state.

regression/pages/common/utils.py
```diff
@@ -0,0 +1,496 @@
...
+
+
+def fill_input_fields(page, selectors_and_values_dict):
+    """
+    A helper function to fill multiple fields in a form
+    Args:
+        page:
+        selectors_and_values_dict:
+    """
+    for key, value in selectors_and_values_dict.iteritems():
```
benpatterson: do you need `iteritems()`?

regression/pages/common/utils.py
```diff
@@ -0,0 +1,496 @@
...
+    """
+    A helper function to click check boxes
+    Args:
+       page:
+       css_selector:
+    """
+    # pylint: disable=unnecessary-lambda
+    page.wait_for_element_visibility(css_selector, 'wait for target checkbox')
+    page_elem = page.q(css=css_selector).results[0]
+    if not page_elem.is_selected():
```
benpatterson: So this will just move on if a checkbox is already checked? I would worry that this would hide a bug.

regression/pages/common/utils.py
```diff
@@ -0,0 +1,496 @@
...
+        page.browser.execute_script("arguments[0].click();", page_elem)
+        page.wait_for(
+            lambda: page_elem.is_selected(),
+            'Target checkbox is selected'
+        )
+
+
+def remove_spaces_from_list_elements(list_with_spaces):
```
benpatterson: I don't understand why you would need this?

regression/pages/common/utils.py
```diff
@@ -0,0 +1,496 @@
...
+    """
+    A helper function to remove spaces from all items in the list
+    Args:
+      list_with_spaces:
+    """
+    return [x.replace(' ', '') for x in list_with_spaces]
+
+
+def get_text_against_page_elements(page, css_selectors):
```
benpatterson: nit: s/against/from

regression/pages/common/utils.py
```diff
@@ -0,0 +1,496 @@
...
+    Args:
+        page:
+        css_selectors:
+    Returns:
+        list_of_text:
+    """
+    list_of_text = []
+    for css_selector in css_selectors:
+        list_of_text.append(page.q(css=css_selector).text[0])
+    return list_of_text
```
benpatterson: I would expect you to return a dict instead of a list, because I'm guessing that the association with the text and where you got the text from is important.

regression/pages/common/utils.py
```diff
@@ -0,0 +1,496 @@
...
+
+
+def extract_mmm_dd_yyyy_date_string_from_text(original_string):
```
benpatterson: Is there a reason you can't just cast a string as a `datetime` object?

regression/pages/common/tests/test_consts.py
```diff
@@ -0,0 +1,311 @@
...
+        'original_date': '2013-1-25',
+        'original_format': 'abcdef',
+        'required_format': '%a %b %d %Y',
+        'result': 'Invalid date or format'
+    }
+}
+
+EXTRACT_NUMERICAL_VALUE = {
+    'normal_string_with_dollar_sign': {
+        "price_strings": '$999',
```
benpatterson: You want other strings here. I see one further below, but there's also `$9999.99`, `$.99`, `$0`, `$99,999`.

regression/pages/common/utils.py
```diff
@@ -0,0 +1,496 @@
...
+        raw_price_string
+    )
+    if regex_result is not None:
+        price_value = regex_result.group(0)
+        return float(price_value.replace(",", ""))
+    else:
+        return 'No numerical value found in search string'
+
+
+def substring_from(original_string, delim):
```
benpatterson: I'm wondering why you wouldn't just say something like `"hi" in "hello hi there"`. I'm not sure why you would want to build a helper function here.

regression/pages/common/utils.py
```diff
@@ -0,0 +1,496 @@
...
+        substring:
+    """
+    if original_string.partition(delim)[1]:
+        return original_string.partition(delim)[1] + \
+               original_string.partition(delim)[2]
+    else:
+        return 'Target substring not found'
+
+
+def get_course_key_from_asset(asset_string):
```
benpatterson: Is this method and the one below it...are they not available in the opaque keys library?

regression/pages/whitelabel/const.py
```diff
@@ -0,0 +1,327 @@
...
+URL_WITHOUT_AUTH = RAW_URL[ORG].format("")
+
+URL_WITH_AUTH = RAW_URL[ORG].format(BASIC_AUTH)
+
+ECOMMERCE_API_URL = ECOMMERCE_URL_WITHOUT_AUTH + 'api/v2/'
+
+CYBERSOURCE_CHECKOUT_URL = \
+    u'https://testsecureacceptance.cybersource.com/checkout'
+
+PROF_COURSE_ID = {
```
benpatterson: Is it ok for us to expose professional course names, IDs in public?

regression/pages/whitelabel/const.py
```diff
@@ -0,0 +1,327 @@
...
+STAFF_EMAIL = os.environ['STAFF_USER_EMAIL']
+
+# Student user email
+STUDENT_USER_EMAIL = 'mit_smoke_test02@yopmail.com'
+
+# Student user email
+VISUAL_USER_EMAIL = 'wl_visual_test01@yopmail.com'
+
+# Countries list
+COUNTRIES_WITH_SPACES = [
```
benpatterson: Is there a reason you can't pull this out of a library somewhere?

regression/pages/whitelabel/const.py
```diff
@@ -0,0 +1,327 @@
...
+    u'Yiddish', u'Yoruba', u'Zhuang', u'Zulu'
+]
+
+LANGUAGES = [x.replace(' ', '') for x in LANGUAGES_WITH_SPACES]
+
+# Default Threshold for visual difference
+DIFF_THRESHOLD = 5000
+
+# Timeouts
+DEFAULT_TIMEOUT = 30
```
benpatterson: These should be further up

regression/pages/whitelabel/const.py
```diff
@@ -0,0 +1,327 @@
...
+    u'Turkmen', u'Turkish', u'Twi', u'Uighur', u'Ukrainian',
+    u'Urdu', u'Uzbek', u'Venda', u'Vietnamese',
+    u'Volap\xfck', u'Welsh', u'Walloon', u'Wolof', u'Xhosa',
+    u'Yiddish', u'Yoruba', u'Zhuang', u'Zulu'
+]
+
+LANGUAGES = [x.replace(' ', '') for x in LANGUAGES_WITH_SPACES]
+
+# Default Threshold for visual difference
+DIFF_THRESHOLD = 5000
```
benpatterson: I see the point but think this is not a smart approach. The diff threshold is based on number of pixels. So, you could have a very small image that is a 0% match, but if it's less than 5000 in size, it will always pass.

regression/pages/common/tests/test_consts.py
kashifch: Yes I guess I missed renaming this, will do for all instances

regression/pages/common/utils.py
kashifch: the emails are generated during the tests so there is a lot of buffer for time zone differences

regression/pages/common/utils.py
kashifch: I have created a cleanup which uses the enrollment apis to un-enroll user(s) from course so that should take care of it. I am removing this

regression/pages/common/utils.py
kashifch: I am looping over the dictionary, if there is a better alternate please let me know.

regression/pages/common/utils.py
kashifch: agreed I will add a check

regression/pages/common/utils.py
kashifch: There is some odd behavior between browsers, when reading the countries lists it seems chrome add some extra spaces in a couple of countries which fails the tests in chrome while passing these in firefox, so I am removing all spaces before comparing the lists

regression/pages/common/utils.py
kashifch: I am just extracting the date from a text chunk, I guess I should rename params and name to make it clear

regression/pages/common/utils.py
kashifch: removing

regression/pages/whitelabel/const.py
kashifch: ok, found a libray

regression/pages/whitelabel/const.py
kashifch: I will change courses with test courses

regression/pages/common/utils.py
benpatterson: Sorry what I meant was, can you accomplish the same thing if you remove `.iteritems()` in this statement.

regression/pages/common/utils.py
benpatterson: I think you should more-fully understand why Chrome is doing this. I think if we create a workaround for poor behavior, it could hide a bug. Do you have any documentation or links describing the problem? I suppose I'm not convinced that we have a bug in our own app here.

regression/pages/whitelabel/const.py
benpatterson: If we need to test actual courses, that is ok. I just think we would remove them from the code and pass it in as a variable (for example).

kashifch:
@benpatterson 
Thanks for all the help, I wasn't able to give it proper time in past few days but I have started the work and I think I will finish all changes and submit these by Monday.

I just need your input on storing sensitive data in environment vars, the tests are currently running on 3 sites, pretty soon there will be 6, we would have multiple courses for all these sites especially to test dynamic coupons. 

using 20-30 environment vars doesn't look like a very good option, is there any other way we could hide the data here? It's just a wild idea but can we put all this data in a password protected private repo and then use it in our tests? or any other way of doing it?

regression/pages/common/utils.py
kashifch: @benpatterson 

I am not sure it is a bug or not as it is not impacting any functionality, however if you open the profile page in firefox and chrome and use inspect element on countries list you will see that in firefox Falkland is shown as Falkland Islands [Malvinas] but in chrome it has 2 spaces between island and [Malvinas], it is like Falkland Islands  [Malvinas].

Do you think I should log a bug about it?

kashifch:
@benpatterson 

Here are most of the modifications.
Please see if the unittests look better.

The countries and languages lists are unchanged as the third party library (pycountry) has too many differences and there is another library but that is django based and we would need to create a django project to use it.

I have removed the EMAIL class and API class from utils as I have added some more apis. I will share it in next PR.

regression/pages/whitelabel/const.py
```diff
@@ -0,0 +1,335 @@
...
+    u'Cayman Islands', u'Central African Republic', u'Chad', u'Chile',
+    u'China', u'Christmas Island',
+    u'Cocos (Keeling) Islands', u'Colombia', u'Comoros', u'Congo',
+    u'Congo (the Democratic Republic of the)',
+    u'Cook Islands', u'Costa Rica', u"C\xf4te d'Ivoire", u'Croatia', u'Cuba',
+    u'Cura\xe7ao', u'Cyprus',
+    u'Czech Republic', u'Denmark', u'Djibouti', u'Dominica',
+    u'Dominican Republic', u'Ecuador', u'Egypt',
+    u'El Salvador', u'Equatorial Guinea', u'Eritrea', u'Estonia',
+    u'Ethiopia', u'Falkland Islands [Malvinas]',
```
benpatterson: I can't find our comments on it, but you had mentioned that Chrome displays this with two spaces before `[Malvinas]` yet firefox with one space. It is cosmetic, but it sounds like a bug. I wouldn't file one yet, though, I would ask a developer for his/her opinion.

benpatterson:
@kashifch Could you explain why we cannot use the same source for countries as we do in production? What are we testing with it?

benpatterson:
I also think the unit tests are more readable now; thank you for moving the data to live with the testcases.

regression/pages/whitelabel/const.py
benpatterson: I think installing a library is much better than a list that you're going to have to maintain and fix over time. (Unless you're attempting to test the library itself, but since it's a 3rd-party library, those kinds of tests should exist on that library.) It is more DRY to re-use that library rather than constructing our own list. It's true that we'll be installing more into the project, but for the most part an installed library is not a big deal.

Taking a step back though, why do you need all the countries?

kashifch: we had some instances in the past when a couple of countries got missed, I think we are doing some customization on django library before displaying it and sometimes during the updates or anything a couple of countries got missed from the list and the issue was notified by white label customers, so we have added a check, in this case we are always alerted whenever there is a change in the list.

Ok I will use the django library and see what steps will be needed in automating the creation of django project plus we would need to do the cutomization that is done in platform, I will just cpy paste the functions from there, but can we do it in a later PR, I have lots of functional changes to focus on and I would rather have some tasks in a to do list.

regression/pages/common/utils.py
kashifch: it gives "too many value to unpack error" without iteritems()

benpatterson:
I don't understand. What do you mean "creation of the django project?"

On platform, what customization do we do of this list?

sent via Android

kashifch:
there you go, removed the lists, we will get back to the library thing later but for now I guess we can dispense with the lists

kashifch:
@benpatterson , done a little more digging and it seems we can use the django countries without creating a project, we just need some setting, here is the piece of code that is creating an exact replica of prod list

from django_countries import countries
from django.conf import settings
from django.core.wsgi import get_wsgi_application

settings.configure()
application = get_wsgi_application()

code_and_name = list(countries)
code_and_name = map(list, zip(*code_and_name))
COUNTRIES = code_and_name[1]
find_index = COUNTRIES.index('Kiribati')
COUNTRIES.insert(find_index + 1, 'Kosovo')

currently the only customization is addition of "Kosovo" in the list.

Also looking at the django countries list I found out that the django countries list has two spaces in Falkland Islands  [Malvinas], so it seems the chrome is displaying the name correctly, firefox is removing one space. 

I will now take a look at the languages

kashifch:
hmm, it seems that languages will pose an issue, in platform a custom list of languages is defined based on http://loc.gov/standards/iso639-2/ISO-639-2_utf-8.txt, the list is defined at lms/env/common.py. It seems that we are using around 186 languages from the original list of around 485.

benpatterson:
@kashifch thank you for digging.

Again, what are you trying to test wherein you need all these countries? Specifically, could you describe the test cases or the functionality that is being tested?

kashifch:
Just checking that all the countries and languages are present in the dropdown of profile page, I don't remember the exact ticket (it was about some countries not loading in the drop down) cause the issue happened a long time ago and Chris Dodge thought it would be good idea to check nothing is missing, if this scenario seems unnecessary to you we can just remove the test case. I haven't seen it failing in the last year except when the spellings for a couple of countries changed and that was legitimate change

kashifch:
do you think this PR is ready to be merged, I will create the next one containing API classes and email class once it is merged

benpatterson:
Ok thanks for the context. 

Checking that nothing is missing....it is overkill to check every single line. You can accomplish the same end by doing a count of the total number of items in the dropdown and making sure it's the right number. 

If you want to be sure the country names are rendered correctly, you could select some of the names (like tricky ones with odd characters) and ensure those are in the dropdown as a separate test case.

I'd go that route. Again, keeping the list of countries up to date will be a maintenance headache and not worth the cost, so the idea is you think of another way of tackling the problem.

kashifch:
apart from the planned scenarios we do add test cases for every scenario that has ever failed on white label sites, like recently the ora file upload was not working on white label so we will be adding a test case for it, also there was an issue that files in static folders were missing the hash after name so we have added a check for it too

kashifch:
hmm ok, I guess we can store the number of countries and languages in a variable plus have a smaller list of some sample items to test for name formats, that way we don't have to keep a big list or install libraries

benpatterson:
it was a bit unclear to me what you were responding to, but, in terms of creating new test cases when a failure is observed, yes that makes sense. again the implementation details of those test cases are where there is room for unmaintainable test code.

kashifch:
sorry for the disconnect but you got the gist of it and I totally agree with you.

On a different note can you take a look at https://build.testeng.edx.org/view/All/job/microsites-staging-tests/ whenever you have some time. I configured it to run using this repo and it worked successfully once but now it is throwing browser errors

benpatterson:
we should take that discussion offline. i'll send an e-mail.

kashifch:
@benpatterson 

I think all the requested changes have been made, if you could give a thumbs up I will merge this and share another PR for review

benpatterson:
👍 from me

benpatterson:
please rebase before merging.

kashifch:
Thanks

PR #89, opened 2016-10-18
Test cleanup

Muddasser:


Muddasser:
@benpatterson @raeeschachar I have added cleanup mechanism in tests. This PR is intended for the confirmation about the procedure and making sure that expectations are met. 

Please review.

raeeschachar:
@Muddasser 
It seems like you have added cleanup for the written tests separately. The idea of adding cleanup is that we have a common function (maybe 2) which cleans up all the studio side and the course is at the default mode for next test/time, regardless of test passes or fails. 
We can have a discussion on this if needed.

regression/tests/studio/test_course_update.py
```diff
@@ -29,47 +29,102 @@ def setUp(self):
...
+                                              ' .post-preview .update-contents'
+        self.course_update_text = 'New update:{}'.format(uuid4().hex)
 
-    def test_course_update(self):
+    def tearDown(self):
         """
-        Verifies creation, editing and deletion of a course update
+        Delete the courses created.
+
+        It is called after every test run irrespective of
```
benpatterson: This particular docstring is not necessary. It's also not entirely accurate; if there is an error, I believe `tearDown` gets skipped; that's one of the reasons I like `addCleanup`.

Muddasser: Interesting. After going through the documentation my impression was that if setUp succeeds,  tearDown will be called whether there is error or not. From [documentation](https://docs.python.org/3.3/library/unittest.html#unittest.TestCase.tearDown) 
"Method called immediately after the test method has been called and the result recorded. This is called even if the test method raised an exception, so the implementation in subclasses may need to be particularly careful about checking internal state. Any exception raised by this method will be considered an error rather than a test failure. This method will only be called if the setUp() succeeds, regardless of the outcome of the test method. The default implementation does nothing."

I can also do it using addCleanup which I think will run all the time even if setUp fails.

Muddasser:
@benpatterson Just to confirm, are we on the same page about @raeeschachar 's suggestion in comment?

benpatterson:
To clarify my position, I think it is ok if we have multiple cleanup functions. Basically the number of cleanup functions should really have more to do with how the code is laid out. Obviously, the more DRY it is the better, but I wouldn't get caught up with the number of functions. 

To me, what's important is that we know the state of the system when we start a test, and to do that, we need to make sure that when we end a test, it's in a known position as well. A bit less important for selenium tests, but we also want to be sure that we are not holding onto any objects or other kinds of memory allocation. That should also happen via cleanup.

benpatterson:
re: `addCleanup` some clarification is [here](https://docs.python.org/3.5/library/unittest.html#unittest.TestCase.addCleanup). Basically `addCleanup` will be run more consistently than `tearDown`, so it's the best practice.

benpatterson:
👍

PR #90, opened 2016-11-02
Tests for Terms and Service and Privacy Policy links

raeeschachar:


raeeschachar:
@benpatterson @muhammad-ammar @Muddasser 
Please review.
I wasn't able to find a better is_browser_on_page for both the pages. Will this work in this case?

regression/pages/studio/privacy_policy.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class PrivacyPolicy(PageObject):
+    """
+    Terms of Service page
+    """
+    url = BASE_URL + '/edx-privacy-policy'
+
+    def is_browser_on_page(self):
+        return 'edX Privacy Policy' in self.browser.title
```
benpatterson: You're right that this is a flaky condition. I dont' know what that page looks like, but I assume there's some privacy policy text? Could you define `is_browser_on_page` on whether or not that particular CSS element is visible?

regression/pages/studio/studio_home.py
```diff
@@ -48,3 +48,17 @@ def click_view_live_button(self):
             "document.querySelectorAll('[data-course-key = \"course-v1:"
             "ArbiRaees+AR-1000+fall\"] .view-button')[0].click();")
         self.browser.switch_to_window(self.browser.window_handles[-1])
+
+    def click_terms_of_service(self):
+        """
+        Clicks Terms of Service link
+        """
+        self.q(css='a[href="https://stage.edx.org/edx-terms-service"]').click()
```
benpatterson: We should use the `BASE_URL` here, too.

regression/pages/studio/terms_of_service.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class TermsOfService(PageObject):
+    """
+    Terms of Service page
+    """
+    url = BASE_URL + '/edx-terms-service'
+
+    def is_browser_on_page(self):
+        return 'edX Terms of Service' in self.browser.title
```
benpatterson: I suggest the same thing here...which is to use whatever CSS element contains the terms of service text, and set it to be visible.

muhammad-ammar:
lgtm 👍

regression/pages/studio/studio_home.py
```diff
@@ -48,3 +48,17 @@ def click_view_live_button(self):
...
+
+    def click_privacy_policy(self):
+        """
+        Clicks Privacy Policy link
+        """
+        self.q(
+            css='a[href="https://stage.edx.org/edx-privacy-policy"]'
```
raeeschachar: Done.

raeeschachar: Same as above.

raeeschachar:
@benpatterson 
Comments addressed. 
I changed to checking the text instead. Is this better? (shouldn't be flaky now)

regression/pages/studio/studio_home.py
benpatterson: My two goals are:
* Reduce repetition (make it DRY)
* Allow the environment to be configurable

So, I suggest a `BASE_STUDIO_URL` or something like that.

raeeschachar:
@benpatterson 
Review?

regression/pages/lms/__init__.py
```diff
@@ -12,3 +12,5 @@
 LOGIN_BASE_URL = 'https://{}:{}@courses.stage.edx.org'.format(
     BASIC_AUTH_USERNAME, BASIC_AUTH_PASSWORD)
 BASE_URL = os.environ.get('test_url', LOGIN_BASE_URL)
+
+BASE_URL_LMS = 'https://stage.edx.org'
```
benpatterson: Ah this looks good, but this is the base URL for the CMS. Could you change it?

raeeschachar: @benpatterson 
Sorry for my confusion, I am not getting this point actually. This is the html link we are going to click:
`self.q(css='a[href="https://stage.edx.org/edx-terms-service"]').click()`
Making 
`BASE_URL_LMS = https://stage.edx.org` will click the link and our test would work but if we change it to CMS `https://studio.stage.edx.org` the click won't be initiated.
What I mean is:
`self.q(css='a[href="https://stage.edx.org/edx-terms-service"]').click()` - This would click
`self.q(css='a[href="https://studio.stage.edx.org/edx-terms-service"]').click()` - This won't and will fail the test.

benpatterson: Ah no you are right. Sorry I saw `stage` but thought I saw `studio`. That's my fault. However, part of my confusion is that `BASE_URL` is not very clear as a standalone variable. Perhaps we need to have `BASE_URL_CMS` and `BASE_URL_LMS`? That would be nice and clear.

That goes beyond the scope of this PR though. Why don't you stick with `BASE_URL_LMS` here, and then put in a separate PR to change `BASE_URL` to `BASE_URL_STUDIO`? That one will take a bit of extra lifting with the testeng jenkins DSL, and we can walk through that, I think.

benpatterson:
Now we're on the same page...thank you for the clarification.
👍 from me.

PR #92, opened 2016-11-02
email and api clients

kashifch:
@benpatterson 

Please take a look whenever you have some time.

We have three different classes for APIs cause it look like different apis hare handled differently, the ecommerce apis could be handled using edx-api-request-client but the for lms apis we need to create a session for logged in user, the enrollment api needs the ACCESS TOKEN of a staff user

### DIFF
--- a/regression/pages/common/api_clients.py
+++ b/regression/pages/common/api_clients.py
@@ -1,0 +1,265 @@
+"""
+API classes for different tasks
+"""
+import Cookie
+import json
+
+import requests
+from requests.auth import AuthBase
+from edx_rest_api_client.client import EdxRestApiClient
+
+from regression.pages.whitelabel.const import (
+    ACCESS_TOKEN,
+    AUTH_PASSWORD,
+    AUTH_USERNAME,
+    ECOMMERCE_API_URL,
+    ENROLLMENT_API_URL,
+    URL_WITHOUT_AUTH
+)
+from regression.pages.common.utils import verify_trailing_slash
+
+
+class ApiException(Exception):
+    """
+    Exceptions for API failures
+    """
+    pass
+
+
+def check_response(response):
+    """
+    Check whether a response was successful. If not, raise an exception
+    Args:
+        response:
+    """
+    if response.status_code != 200:
+        raise ApiException(
+            'API request failed with following error code: ' +
+            str(response.status_code) + '\n' + response.text
+        )
+
+
+class BearerAuth(AuthBase):
+    """
+    Attaches Bearer Authentication to the given Request object.
+    """
+
+    def __init__(self, token):
+        """
+        Instantiate the auth class.
+        """
+        self.token = token
+
+    def __call__(self, request):
+        """
+        Update the request headers.
+        """
+        request.headers['Authorization'] = 'Bearer {}'.format(self.token)
+        return request
+
+
+class EcommerceApiClient(object):
+    """
+    Ecommerce API client for various actions like creating, deleting
+    downloading coupons
+    """
+
+    @property
+    def ecommerce_api_client(self):
+        """
+        E-Commerce API Client
+        """
+        return EdxRestApiClient(
+            ECOMMERCE_API_URL, oauth_access_token=ACCESS_TOKEN)
+
+    def create_coupon(self, payload):
+        """
+        Create coupons using API
+        Args:
+            payload:
+        """
+        response = self.ecommerce_api_client.coupons.post(data=payload)
+        if 'coupon_id' not in response.keys():
+            raise ApiException('Coupon not returned')
+        return response['coupon_id']
+
+    def get_coupon_report(self, coupon_id):
+        """
+        Get coupon report using api
+        Args:
+            coupon_id:
+        """
+        response = self.ecommerce_api_client.products(coupon_id).get()
+        if not response:
+            raise ApiException('No coupon report found')
+        return response
+
+    def delete_coupon(self, coupon_id):
+        """
+        Delete the coupon created in the tests and all associated stock
+        records and vouchers.
+        Args:
+            coupon_id:
+        """
+        self.ecommerce_api_client.coupons(coupon_id).delete()
+
+    def get_coupon_codes(self, coupon_id):
+        """
+        Get coupon codes from coupon report
+        Args:
+            coupon_id:
+        """
+        coupon_codes = []
+        items = (
+            self.get_coupon_report(coupon_id)['attribute_values'][0]['value']
+        )
+        for item in items:
+            coupon_codes.append(item['code'])
+        return coupon_codes
+
+
+class EnrollmentApiClient(object):
+    """
+    Enrollment API client
+    """
+
+    def __init__(self, host=None, key=None):
+        self.host = verify_trailing_slash(host or ENROLLMENT_API_URL)
+        self.key = key or ACCESS_TOKEN
+
+    def is_user_enrolled(self, username, course_id):
+        """
+        Retrieve the enrollment status for given user in a given course.
+        Args:
+            username:
+            course_id:
+        Returns:
+            True or false based on whether user is enrolled
+        """
+        url = '{host}enrollment/{username},{course_id}'.format(
+            host=self.host,
+            username=username,
+            course_id=course_id
+        )
+        response = requests.get(url, auth=BearerAuth(self.key))
+        check_response(response)
+        formatted_response = response.json()
+        return formatted_response['is_active']
+
+
+class LmsApiClient(object):
+    """
+    API class for accessing different APIs
+    """
+
+    def __init__(self, host=None):
+        """
+        Instantiate the api class.
+        """
+        self.host = verify_trailing_slash(host or URL_WITHOUT_AUTH)
+        self.session = requests.Session()
+        self.session.auth = (AUTH_USERNAME, AUTH_PASSWORD)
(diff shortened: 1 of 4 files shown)


kashifch:
@benpatterson please take a look when you find some time, also it seems that there is a port issue on jenkins that is not allowing my scripts to access gmail. Since it was working correctly in the past there may be some enhanced security settings or something which is causing the blockage. please do take a look when you have some time

regression/pages/common/api_clients.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+def check_response(response):
+    """
+    Check whether a response was successful. If not, raise an exception
+    Args:
+        response:
+    """
+    if response.status_code != 200:
+        raise ApiException(
```
benpatterson: you may want to include response.text as well.

regression/pages/common/api_clients.py
```diff
@@ -0,0 +1,260 @@
...
+    Attaches Bearer Authentication to the given Request object.
+    """
+
+    def __init__(self, token):
+        """ Instantiate the auth class. """
+        self.token = token
+
+    def __call__(self, r):
+        """ Update the request headers. """
+        r.headers['Authorization'] = 'Bearer {}'.format(self.token)
```
benpatterson: I seem to remember that we don't have pylint on this repo? One-letter variables are not the best...could you please re-name?

regression/pages/common/api_clients.py
```diff
@@ -0,0 +1,260 @@
...
+        return EdxRestApiClient(
+            ECOMMERCE_API_URL, oauth_access_token=ACCESS_TOKEN)
+
+    def create_coupon(self, payload):
+        """
+        Create coupons using API
+        Args:
+            payload:
+        """
+        response = self.ecommerce_api_client.coupons.post(data=payload)
```
benpatterson: Do you need headers in here? Do those come for free with the client?

regression/pages/common/api_clients.py
```diff
@@ -0,0 +1,260 @@
...
+        if 'coupon_id' not in response.keys():
+            raise ApiException('Coupon not created')
```
benpatterson: It might be more accurate to say "Coupon not returned."

regression/pages/common/api_clients.py
```diff
@@ -0,0 +1,260 @@
...
+        return response['coupon_id']
+
+    def get_coupon_report(self, coupon_id):
+        """
+        Get coupon report using api
+        Args:
+            coupon_id:
+        """
+        response = self.ecommerce_api_client.products(coupon_id).get()
+        if not response:
```
benpatterson: Is this better as a try/except? Would we really get a `None` object if there was a failure in a `get`?

regression/pages/common/api_clients.py
```diff
@@ -0,0 +1,260 @@
...
+
+    def delete_coupon(self, coupon_id):
+        """
+        Delete the coupon created in the tests and all associated stock
+        records and vouchers.
+        Args:
+            coupon_id:
+        """
+        response = self.ecommerce_api_client.coupons(coupon_id).delete()
+        if not response:
```
benpatterson: Same question here....will an exception be raised if something goes wrong on the `.delete()` call? I really don't know...it's likely in the client but I wanted to ask.

regression/pages/common/api_clients.py
```diff
@@ -0,0 +1,260 @@
...
+        Args:
+            coupon_id:
+        """
+        coupon_codes = []
+        items = (
+            self.get_coupon_report(coupon_id)['attribute_values'][0]['value']
+        )
+        for item in items:
+            coupon_codes.append(item['code'])
+        return coupon_codes
```
benpatterson: Will you see repeated codes in the report? Is that something you want to worry about in the list that you return?

regression/pages/common/api_clients.py
```diff
@@ -0,0 +1,260 @@
...
+        Args:
+            user_email:
+            user_password:
+        """
+        self.create_base_session()
+        login_url = self.host + 'user_api/v1/account/login_session/'
+        payload = {
+            'email': user_email,
+            'password': user_password,
+            'remember': 'false'
```
benpatterson: What's `remember` do?

regression/pages/common/api_clients.py
```diff
@@ -0,0 +1,260 @@
...
+        }
+        response = self.session.post(login_url, data=payload)
+        check_response(response)
+        self.session.cookies = response.cookies
+        self.session.headers = self._post_headers(
+            response.cookies['csrftoken']
+        )
+        self.login_response = response
```
benpatterson: Do you want this method to handle or return any errors or failures?

regression/pages/common/api_clients.py
```diff
@@ -0,0 +1,260 @@
...
+        return user_info_dict['username']
+
+    def change_enrollment(self, course_id, enrollment_action):
+        """
+        Change enrollment
+        Args:
+            course_id:
+            enrollment_action:
+        """
+        change_enrollment_url = self.host + 'change_enrollment'
```
benpatterson: This is going to assume that `self.host` ends with a slash. You may want to handle that if it's not the case.

kashifch:
The activation, password change and bulk purchasing all send email to users (old user and new ones). Like in case of activation I need to have a valid email address for registration and after registration I need to go to that inbox and read the mail user has received and get the activation link from it. So I have created a gmail account for testing and for all new users I use an alias, I use the python imap to connect to gmail inbox.

Previously I was using tempmail api, where I could create a temporary account on tempmail and read the inbox using the api but the problem was  that that these free email servers proved quite inconsistent. If there is a way that I can check/receive the emails generated from the system without using any thrid party interference it would be great

regression/pages/common/api_clients.py
kashifch: done

regression/pages/common/api_clients.py
kashifch: no need for headers when using edxRestApiClient

regression/pages/common/api_clients.py
kashifch: it is the value for remember me check box below the login

regression/pages/common/api_clients.py
kashifch: Client only returns a dictionary of values and does not return any status code or something so I could not check the 200 or ok status. In case of any errors the client will raise a slumber http exception which i don't think we need to catch, the only eventuality I am covering here is that if client returns an empty list or dictionary. sometimes the coupon ids can exist without any valid data attached with these so in that scenario I would like to raise an exception

regression/pages/common/api_clients.py
kashifch: removing the return here, if anything goes wrong the slumber http exception will be generated

regression/pages/common/api_clients.py
kashifch: nope, the codes are unique

regression/pages/common/api_clients.py
kashifch: the check_response function is checking for errors

regression/pages/common/api_clients.py
kashifch: all our base urls end with a slash so it shouldn't be an issue

regression/pages/common/api_clients.py
benpatterson: In the case of a non-200 code though, I would expect there to be less text. My main point is to think through what will help someone who is troubleshooting the issue. If you think the text is going to be more noise than it's worth, then that's ok. You can stick with what you have. But pls look at a couple of examples of failures before you decide.

regression/pages/common/api_clients.py
benpatterson: Ok

regression/pages/common/api_clients.py
benpatterson: Ok I understand. `products().get()` would return `None`?

regression/pages/common/api_clients.py
benpatterson: Ok. A comment would help in this case.

regression/pages/common/api_clients.py
benpatterson: They end with a slash *now* but someone may make a change later in which they don't. Or will set an environment variable where they don't. So you just want to think through that and assume someone makes a manual mistake somewhere. (If it were me, I would not know if I need a slash or not, so it's a bit arbitrary what I might set locally.)

You need either instructions to end with a slash, or (better) a utility that ensures `self.host` ends with a slash.

benpatterson:
@kashifch Could you enter an IT ticket to create test users for you with email addresses at edX that you can control? It'll just be a bit better to make the test users something that can be centrally maintained.

kashifch:
@benpatterson I like the idea, I think we would need just one user, how about e2etests@edx.org, we can just use aliases to derive more users

benpatterson:
Great. Could you put in an IT ticket please?

Muddasser:

kashifch:
@benpatterson, @raeeschachar, @Muddasser 

Please review this branch, I have removed the email client part so it is just the api client for now. I have incorporated al the feedback as well. Once it is merged I will create a new branch for the next chunk of code

feanil:
Closing due to age and lack of activity.

PR #94, opened 2016-11-15
content visibility help

Muddasser:


Muddasser:
@raeeschachar @benpatterson Please review

regression/tests/studio/test_course_outline.py
```diff
@@ -53,3 +55,28 @@ def test_edit_start_button(self):
...
+    def test_content_visibility_help(self):
+        """
+        Scenario: Help link for
+        'Learn more about content visibility settings' is working
+        Given that I am on the 'Course Outline' page.
+        And I want help about the content visibility settings.
+        And I click 'Learn more about content visibility settings'
+        in the sidebar
+        Then Help link should open.
+        And help url should contain 'developing_course/
```
benpatterson: nit: I think you can make this docstring more generic. (So that if we change the link, you only need to change the code, not the docstring.) You can say something like "And url should point to the content visibility help page"

benpatterson:
A nit, but 👍 whether you decide to update the docstring or not. Your call.

raeeschachar:
👍

PR #95, opened 2016-11-17
Test Other Links working on Schedule and Details page

raeeschachar:


raeeschachar:
Strange that locally import order error doesn't appear on Pylint check but fails here, anyways

@benpatterson @Muddasser 
Please review.

benpatterson:
@raeeschachar re: the pylint error... it might depend on which directory you were in when you ran pylint vs what the build does....

regression/tests/studio/test_schedule_and_details_settings.py
```diff
@@ -94,3 +102,70 @@ def test_cancel_upload(self):
...
+            self.course_info['number'],
+            self.course_info['run']
+        )
+        self.settings_page.visit()
+
+    def test_other_grading_link(self):
+        """
+        Verifies that user can click and navigate to Grading
+        """
+        link = 'href="/settings/grading/course-v1:ArbiRaees+AR-1000+fall"'
```
benpatterson: This is another case where---on a different environment---tests would fail. I wonder if you can pull the course out as another variable... Or, if we want to think of another, scalable way to handle these situations. Perhaps we need to create courses as part of test setup? Or, perhaps we want a dedicated test data file (like a *.ini file) that we include when running tests?

regression/tests/studio/test_schedule_and_details_settings.py
```diff
@@ -94,3 +102,70 @@ def test_cancel_upload(self):
...
+            self.course_info['number'],
+            self.course_info['run'])
+        self.settings_page.click_other_settings_links(link)
+        grading_page.wait_for_page()
+
+    def test_other_course_team_link(self):
+        """
+        Verifies that user can click and navigate to Course Team
+        """
+        link = 'href="/course_team/course-v1:ArbiRaees+AR-1000+fall"'
```
benpatterson: At the very least, since it's the same course, you should have this as a var on the top of the file.

Muddasser:

raeeschachar:
@benpatterson @Muddasser 
I think I understand what you were trying to explain. We don't really need href for the css selector, we can do it by giving index in .results. I have restructured the test, let me know if this works.

benpatterson:
@Muddasser it looks like the jenkins/e2e context is run on your machine? Is there a reason we don't have it on our Jenkins instance yet?

raeeschachar:
@Muddasser @benpatterson 
Please review.

benpatterson:
I didn't notice this before, but you're passing the `click_other_settings_links` as an index. It is better to send it a link title, or something more descriptive. That way, if the order of the links changes, the test does not break. Besides, it is difficult to tell what the destination of the link is supposed to be when reading a '0' or a '1', etc.

raeeschachar:
@benpatterson @Muddasser 
Done. Please review.

Muddasser:
@raeeschachar 👍 

@benpatterson I don't thing there is any reason. Let me talk to Stuart today to get insights.

PR #96, opened 2016-11-18
Verification of added textbook on LMS

raeeschachar:


raeeschachar:
@benpatterson  @Muddasser 
Please review.

regression/pages/lms/lms_textbook.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class TextbookPage(PageObject):
+    """
+    Textbook page LMS
+    """
+    url = None
+
+    def is_browser_on_page(self):
+        return self.q(css='.chapter').present
```
benpatterson: Should this be `.visible`?

regression/pages/studio/studio_textbooks.py
```diff
@@ -69,3 +70,14 @@ def delete_all_textbooks(self):
         """
         while self.get_textbook_count() > 0:
             self.click_delete_button()
+
+    def click_view_live_textbook(self):
+        """
+        Clicks View Live button on the first textbook available
+        """
+        button = self.q(css='.view').results[0]
+        # This button is hidden, hovering on it makes it visible
```
benpatterson: Don't need to change the test, but I wonder if this is considered an accessibility issue. Did you happen to notice if you could tab to it?

raeeschachar: We can tab to it, but it doesn't display.

raeeschachar:
@benpatterson  @Muddasser 
Comments addressed. Review?

regression/pages/studio/studio_textbooks.py
benpatterson: That's ok. Once someone tabs to it, can they hit enter and un-hide it? That would help folks that need to use a keyboard, and it might imply that folks with a screen reader can also read the text.

benpatterson:
I still have some clarifying questions but the code itself is ready to merge from my perspective. Certainly get feedback from others (i.e., I should not be a bottleneck. :) )

regression/pages/studio/studio_textbooks.py
raeeschachar: Understood. Yes they can hit Enter and it does go to where expected.

Muddasser:
👍

Muddasser:

PR #97, opened 2016-11-18
drap/drop and hide/show

Muddasser:


Muddasser:
This PR is for [LT-32](https://openedx.atlassian.net/browse/LT-32) and [LT-34](https://openedx.atlassian.net/browse/LT-34)

@raeeschachar @benpatterson Please review.

regression/pages/lms/course_page_lms.py
```diff
@@ -22,3 +22,13 @@ def click_resume_button(self):
         Clicks Resume button of the course selected
         """
         self.q(css='.last-accessed-link').first.click()
+
+    def get_pages_name(self):
```
benpatterson: Interesting. I think you've put this change up before, but would you mind sending me an example of the results? Will it also give you the sequences within each unit, for example?

regression/pages/studio/pages_page_studio.py
```diff
@@ -143,3 +144,67 @@ def click_and_verify_see_an_example(self):
...
+    def click_hide_show_toggle(self, index=0):
+        """
+        Clicks hide/show toggle button
+        """
+        toggle_checkbox_css = '.toggle-checkbox'
+        toggle_checkbox = self.q(css=toggle_checkbox_css).results[index]
+        ActionChains(self.browser).move_to_element(
+            toggle_checkbox
+        ).click().perform()
+        wait_for_notification(self)
```
benpatterson: What notification are you waiting for here?

regression/pages/studio/pages_page_studio.py
```diff
@@ -143,3 +144,67 @@ def click_and_verify_see_an_example(self):
...
+        return self.browser.execute_script(
```
benpatterson: Could you explain why you need to get this done via javascript?

regression/pages/studio/pages_page_studio.py
```diff
@@ -143,3 +144,67 @@ def click_and_verify_see_an_example(self):
...
+            "return $('{}:eq({})').parents().eq(3).find("
+            "'.course-nav-item-header .title').text()".format(
+                toggle_checkbox_css, index
+            )
+        )
+
+    def get_movable_default_pages_order(self):
```
benpatterson: nit: I'd put this method below `get_moveable_default_pages`

regression/pages/studio/pages_page_studio.py
```diff
@@ -143,3 +144,67 @@ def click_and_verify_see_an_example(self):
...
+        """
+        Returns the order of moveable default pages
+        """
+        return self.q(
+            css='.course-nav-item.is-movable .course-nav-item-header'
+        ).text
```
benpatterson: It's odd it would return text. Shouldn't it be a list or other iterable?

regression/pages/studio/pages_page_studio.py
```diff
@@ -143,3 +144,67 @@ def click_and_verify_see_an_example(self):
...
+
+    def get_moveable_default_pages(self):
+        """
+        Returns pages which are default and moveable
+        """
+        return self.q(
+            css='.course-nav-item.is-movable .drag-handle.action'
+        ).results
```
benpatterson: Same here, I'm wondering if the result should be a list. Is that' what's returned?

regression/pages/studio/pages_page_studio.py
```diff
@@ -143,3 +144,67 @@ def click_and_verify_see_an_example(self):
...
+
+    def is_page_shown(self, index=0):
+        """
+        Check whether a page is shown or not at a given index
```
benpatterson: I'm confused...I don't see any `wait_for` kinds of methods in here. When you say "whether a page is shown or not" do you really mean "whether a page is *configured to* show or not?"

regression/pages/studio/pages_page_studio.py
```diff
@@ -143,3 +144,67 @@ def click_and_verify_see_an_example(self):
...
+            bool: True if shown otherwise False.
+        """
+        toggle_value = self.q(
+            css='.toggle-checkbox'
+        ).results[index].get_attribute('checked')
+        if toggle_value:
+            return False
+        return True
+
+    def drag_and_drop(self, source, target):
```
benpatterson: I think we have this method in our utils in the edx-platform bok-choy files. Could you re-use that?

regression/tests/studio/studio_base_test.py
```diff
@@ -80,6 +80,9 @@ def clean_up_function(self):
             pages_page.wait_for_the_visibility_of_new_page()
             pages_page.delete_all_pages()
 
+        if pages_page.is_page_shown() is False:
```
benpatterson: I don't think we should put this here. This method basically says "if it's not showing, just click the toggle button". That's smart programming, but for a test, you should know the expected state of the page. If it's not showing--but it should be...then you should encounter an error. 

Putting this kind of logic here could hide display bugs.

regression/tests/studio/test_course_pages.py
```diff
@@ -148,3 +149,97 @@ def test_view_live_pages(self):
...
+        self.pages_page.drag_and_drop(
+            moveable_pages[-1],
+            moveable_pages[-2]
+        )
+
+        temp_page = moveable_pages_order[-2]
+        moveable_pages_order[-2] = moveable_pages_order[-1]
+        moveable_pages_order[-1] = temp_page
+
+        # Asset that order of pages has been changed.
```
benpatterson: nit:s/Asset/Assert

regression/tests/studio/test_course_pages.py
```diff
@@ -148,3 +149,97 @@ def test_view_live_pages(self):
...
+        self.assertEqual(
+            self.pages_page.get_movable_default_pages_order(),
```
benpatterson: Ah, in this case, perhaps we shouldn't call the method `default` pages order. Maybe it should just be `get_moveable_pages_order`

regression/tests/studio/test_course_pages.py
```diff
@@ -148,3 +149,97 @@ def test_view_live_pages(self):
...
+        course_info = get_course_info()
+        course_page = CourseInfoPageExtended(
+            self.browser, get_course_key(course_info)
+        )
+        course_page.visit()
+        pages_in_tab = course_page.get_pages_name()
+
+        # Convert lists pages_in_tab and moveable_pages_order to string.
+        # It makes it easier to assert the order.
+        string_of_tab_pages = ''.join(pages_in_tab)
```
benpatterson: I think we have assertions for lists...I don't think you need to do this. You can ask on hipchat for help.

regression/tests/studio/test_course_pages.py
```diff
@@ -148,3 +149,97 @@ def test_view_live_pages(self):
...
+        course_info = get_course_info()
+        course_page = CourseInfoPageExtended(
+            self.browser, get_course_key(course_info)
+        )
+
+        course_page.visit()
+        pages_in_tab = course_page.get_pages_name()
+        self.assertIn(page_name, pages_in_tab)
+
+    def assert_page_is_not_shown_in_lms(self, page_name):
```
benpatterson: negative assertion. awesome!

regression/pages/lms/course_page_lms.py
Muddasser: It will output something like this [u'Home\n, current location', u'Course', u'Discussion', u'Wiki', u'Progress', u'Instructor']. It will return pages in sequence. I am sorry I don't have any clue about putting it before. Its in LMS. Can you pin point?

regression/pages/studio/pages_page_studio.py
Muddasser: It waits for the notification at the bottom right of the page.

regression/pages/studio/pages_page_studio.py
Muddasser: Well it was a last minute whim. The query was complex, and I thought it would be better to do it directly in jQuery.

regression/pages/studio/pages_page_studio.py
Muddasser: True, it is a list.

regression/pages/studio/pages_page_studio.py
Muddasser: Good suggestion. Changing the wordings of docstring.

regression/tests/studio/test_course_pages.py
Muddasser: Ah, I guess we need to refactor it a bit. Specifically, to use list assertion, I am considering refining the list returned by get_pages_name.

regression/pages/lms/course_page_lms.py
benpatterson: Ok, I think this is moreso giving you the list of tabs. Should probably clarify that it's tabs and not individual pages (which appear in a sequence). Please rename the method?

regression/pages/studio/pages_page_studio.py
benpatterson: Please clarify in the docstring that it's returning a list

regression/pages/studio/pages_page_studio.py
benpatterson: Same thing...just clear up any confusion in the docstring

Muddasser:
@benpatterson Some refactoring. Please have a view.

regression/pages/lms/course_page_lms.py
```diff
@@ -31,4 +31,5 @@ def get_pages_name(self):
         Returns
             list: A list of names of all pages
         """
-        return self.q(css='.tabs.course-tabs .tab').text
+        tab_pages = self.q(css='.tabs.course-tabs .tab').text
+        return [text.replace('\n, current location', "") for text in tab_pages]
```
benpatterson: Could you explain why we're adding this text? Wouldn't the index in the list be good enough?

regression/pages/studio/pages_page_studio.py
```diff
@@ -162,25 +161,28 @@ def click_hide_show_toggle(self, index=0):
...
-        return self.q(
-            css='.course-nav-item.is-movable .drag-handle.action'
-        ).results
+        return len(
+            self.q(
+                css='.course-tab'
+            ).results
+        )
 
     def is_page_shown(self, index=0):
```
benpatterson: Could you rename the method to something like `is_page_configured_to_show` or `is_show_page_set`.. something like that?

benpatterson:
Good improvements; I had a couple more comments if you could take a look.

regression/pages/lms/course_page_lms.py
Muddasser: There is an extra text 'Current location' alongside the page's name of selected(active) tab. If we don't remove it then it would pose problems in assertions latter on.

raeeschachar:
👍

PR #99, opened 2016-11-24
Remember me functionality on Login page

raeeschachar:


raeeschachar:
@benpatterson @Muddasser 
Please review. 
[Bug](https://openedx.atlassian.net/browse/TNL-6008) was added. Sending the PR so the code gets reviewed and once the Bug is fixed, we can edit and merge.

regression/pages/lms/dashboard_lms.py
```diff
@@ -29,3 +29,13 @@ def click_donate_button(self):
         Clicks donate button on Dashboard
         """
         self.q(css='.action-donate').click()
+
+    def logout_lms(self):
```
benpatterson: nit: I'd call this `click_logout_lms` because you're doing the `wait` separately (in fact, one of the points of your test.)

raeeschachar:
@benpatterson @Muddasser 
Changed the test to verify cookies instead. Please review.

benpatterson:
Yep that makes sense. You may want to extract the name of the cookie as a variable so you're not writing it out 3 times. But overall LGTM. 👍

regression/pages/lms/lms_home_page.py
```diff
@@ -0,0 +1,18 @@
+"""
+LMS Home page
+"""
+from bok_choy.page_object import PageObject
+from regression.pages.lms import BASE_URL
+
+
+class LmsHome(PageObject):
```
Muddasser: Can we extend it from https://github.com/edx/edx-platform/blob/master/common/test/acceptance/pages/lms/index.py ?

Muddasser:
Just a small change. 👍

regression/pages/lms/lms_home_page.py
raeeschachar: @Muddasser 
We tend to keep it separate because we are using stage and we won't be able to use most of the methods. I think that is what we decided. Do you want me to change it?

PR #100, opened 2016-11-30
Pagination test on files and uploads page

raeeschachar:


raeeschachar:
@benpatterson @Muddasser @muhammad-ammar 
Please review.

regression/pages/studio/asset_index_studio.py
```diff
@@ -53,23 +52,17 @@ def get_files_count(self):
...
-        """
-        more_then_one_file = False
-        for index in file_indexes:
-            if more_then_one_file and index != 0:
-                index = index - 1
-            click_css_with_animation_enabled(
-                self, '.remove-asset-button.action-button', index, False)
-            confirm_prompt_with_animation_enabled(
-                self, require_notification=False)
+        Deletes the file till delete asset button is found on page
```
benpatterson: This is another instance of: sensible programming, bad test pattern. The tests should be built in such a way that we know how many files should be appearing, and we know how many times we should be clicking. Otherwise, we may hide bugs on this action. I should have noticed this on the original implementation as well. 

It could be a "less smart" function: click the delete button. Then in the test, you do a wait to see if the file has been deleted. If you have several files, you would cycle through that on the test, rather than the helper function. And you'd do it a known amount of times. I hope this feedback makes sense?

regression/tests/studio/studio_base_test.py
```diff
@@ -94,5 +94,4 @@ def clean_up_function(self):
         textbook_page.delete_all_textbooks()
 
         asset_page.visit()
-        asset_count = asset_page.get_files_count()
-        asset_page.delete_file(list(range(asset_count)))
+        asset_page.delete_file()
```
benpatterson: So under this line, you'd see if the file has been deleted (possibly in a `wait_for` statement, but not necessarily...)

raeeschachar:
@benpatterson @Muddasser 
I think I understand what you meant. Are you expecting something like this? Please review.

regression/pages/studio/asset_index_studio.py
```diff
@@ -111,3 +104,17 @@ def sort_assets(self):
         Sort the assets
         """
         click_css_with_animation_enabled(self, '.column-sort-link', 0, False)
+
+    def get_page_count(self):
```
MuddasserHussain: It returns page count as string. Either you can return a numerical value or can change the docstring. I know what you are trying to say, but its little ambiguous at the moment.

Muddasser:
👍

benpatterson:
@raeeschachar It is still going to keep deleting until it gets to zero. Let's say, for example, that there was a bug and the delete button didn't work. The count would say >0 and it would just keep hitting delete infinitely. 

Instead the test should know how many things there are to delete, and delete only that many. Does that make sense?

raeeschachar:
@benpatterson 
Please review.

benpatterson:
👍

PR #103, opened 2016-12-19
Fixes course issue. Added Demo course for permanent use

raeeschachar:


raeeschachar:
@benpatterson @Muddasser 
This test has been failing due to a couple of reasons:
1- The course we were using was no longer available on stage. 
2- CyberSource displays http://screencast.com/t/ck8BL1GBXHA

I have fixed the first problem by using Demo course. I am not sure if we have any solution for the second one as it is a third party tool we are using. Any suggestions?

benpatterson:
The cybersource problem is important and there might not be a quick solution. Could you track down who owns that integration? I'd start with @rlucioni . He may not own it, but he'd know where to go...

rlucioni:
I remember @mjfrey was seeing that CyberSource page on stage last Friday. @mjfrey has this issue been resolved?

raeeschachar:
@cpennington FYI

mjfrey:
The new client side checkout page is live on stage and course titles can only contain alpha numeric values as cybersource will not like the signature with anything else.

raeeschachar:
@jzoldak @benpatterson @Muddasser 
Please review. Changed this PR according to [this ticket](https://openedx.atlassian.net/browse/LT-48)

benpatterson:
@raeeschachar is this still in progress? Looks like the donations test fails; I wasn't sure if solving that failure was still the goal.

raeeschachar:
jenkins run e2e

raeeschachar:
jenkins run e2e

raeeschachar:
jenkins run e2e

raeeschachar:
Test passes. https://screencast.com/t/VXknTIGcgK1d
Good to merge?

MuddasserHussain:
@raeeschachar Looks alright to me.

Muddasser:
@raeeschachar Looks alright to me.

jzoldak:
image
@raeeschachar I am still able to reproduce the flakiness locally. I have a hypothesis what it is, I've attached a screenshot illustrating. My guess is that the user is getting to the dashboard too quickly - before the donation header appears. And then waiting for it is useless because it will never appear. But it does seem to be drawn when the page is reloaded in the browser.

So something else needs to happen within/between/after the  [click_enroll_now](https://github.com/edx/edx-e2e-tests/blob/master/regression/tests/lms/test_donations.py#L53) and dashboard_page.wait_for_page call in this piece of code: 
```
        self.drupal_course_page.visit()
        self.drupal_course_page.click_enroll_now()
        self.dashboard_page.wait_for_page()
        self.dashboard_page.click_donate_button()
```

Couple suggestions off the top of my head. Use your judgement:
* modify the click_enroll_now to do some sort of synchronization (I don't know if this makes sense)
* after or instead of the dashboard_page.wait_for_page call, do a visit to the dashboard page

raeeschachar:
@jzoldak 
Done.

jzoldak:
@raeeschachar LGTM 👍

PR #104, opened 2016-12-22
Login page test fix

raeeschachar:


raeeschachar:
@benpatterson @Muddasser 
Fixes login page test that was failing on edx-e2e-tests. Please review. 
cc: @cpennington

PR #110, opened 2017-01-03
Tests will now support concurrent run

raeeschachar:


raeeschachar:
@benpatterson @Muddasser 
Please review. 
Idea here is that when setup is called (for every test), a new course rerun is made and all the tasks are performed on that particular rerun, as discussed.

regression/pages/studio/studio_home.py
```diff
@@ -62,3 +62,11 @@ def click_privacy_policy(self):
         """
         self.q(
             css='a[href="' + BASE_URL_LMS + '/edx-privacy-policy"]').click()
+
+    def click_course_rerun(self):
+        """
+        Clicks rerun course button
+        """
+        self.browser.execute_script(
```
benpatterson: I'm curious why you need to get this done via javascript.

regression/pages/studio/studio_home.py
```diff
@@ -62,3 +62,11 @@ def click_privacy_policy(self):
...
+            "document.querySelectorAll('[data-course-key = \"course-v1:"
+            "ArbiRaees+AR-1000+fall\"] .rerun-button')[0].click();")
```
benpatterson: Should the course key be a var?

regression/pages/studio/studio_rerun.py
```diff
@@ -0,0 +1,28 @@
...
+from uuid import uuid4
+from bok_choy.page_object import PageObject
+from regression.pages.studio import BASE_URL
+
+
+class StudioRerun(PageObject):
+    """
+    Studio Course Rerun for Manual Smoke Test Course
+    """
+    url = BASE_URL + '/course_rerun/course-v1:ArbiRaees+AR-1000+fall'
```
benpatterson: Should the course name be a var here as well?

benpatterson: IIRC it is a parameter we pass into the Jenkins job

regression/tests/studio/studio_base_test.py
```diff
@@ -23,6 +26,22 @@ class StudioBaseTestClass(WebAppTest):
...
+        self.login_page = StudioLogin(self.browser)
+        self.course_rerun_page = StudioRerun(self.browser)
+        self.studio_home_page = DashboardPageExtended(self.browser)
+        LoginHelper.login(self.login_page)
+        self.studio_home_page.visit()
+        self.studio_home_page.click_course_rerun()
+        self.course_rerun_page.wait_for_page()
+        self.course_rerun_page.add_rerun_with_run()
+        self.studio_home_page.wait_for_page()
+        self.browser.refresh()
```
benpatterson: why the refresh here?

raeeschachar:
jenkins run e2e

regression/pages/studio/studio_home.py
raeeschachar: @benpatterson 
Button appears after hovering on it, I tried using move_to_element but the tests were flaky. So used JS to make it more reliable.

regression/tests/studio/studio_base_test.py
raeeschachar: @benpatterson 
After rerun is added [this appears](http://screencast.com/t/Rxh8FTVy8H), which requires refresh.

raeeschachar:
@benpatterson @jzoldak 
Comments addressed. Please review. 
After this PR is merged, tests take about [54 mins](https://build.testeng.edx.org/job/edx-e2e-tests/100/testReport/) with 2 failures. (On this we can run tests concurrently)
Before this tests took [38 mins](https://build.testeng.edx.org/job/edx-e2e-tests/99/testReport/)
Suggestions?

raeeschachar:
Reminder. Review? :)

benpatterson:
​I wonder if we can multithread with this change?

Do you know how to run a profile on a test? That might prove interesting.

jzoldak:
@raeeschachar sorry for not leaving feedback.
I think that there are some problems with this approach that outweigh the benefits. 
* The biggest one, like you mentioned, tests now take a lot longer to run. 
* Also, these new course enrollments are piling up on the user's dashboard, making that page take longer and longer to render.
* Not a hard requirement, but I would like to avoid programmatically creating and leaving course runs on staging if possible. We could come up with a janitor script to delete them if we need to, but I'd rather not if not needed.

Another approach that might work better - let's look at the tests holistically, understand what each is trying to verify, and rethink which existing course might serve those needs. Then mark tests with nose attributes (or pytest markers if we want to switch from nose to pytest) and run them in shards like we do the bok-choy tests, with the shards able to run at the same time because they touch different courses.

Hope this makes sense. I'm also doing some experiments on a branch, so if I get anything helpful going I'll put it in as a PR and tag you, even if it's just a WIP.

raeeschachar:
@jzoldak 
Thank you for the input. We had in mind that devops would be regularly removing courses from stage so @benpatterson @Muddasser and me thought of changing the tests so that they are not dependent on each other. 
I think this is a nice idea too, I will go ahead and start working on the provided suggestions and will keep my eye on your PRs too.

benpatterson:
Andy pointed out this page on deleting courses. It can only be done via management command. If we create the jenkins dsl for it, I suspect devops would roll it out for us.

benpatterson:
sorry the page is: https://openedx.atlassian.net/wiki/display/PLAT/How+to+Delete+a+Course

jzoldak:
If we go with a course-deleting management command, we should remember that this does **not** remove course enrollments, and we should programmatically remove them also.

raeeschachar:
@benpatterson 
https://openedx.atlassian.net/wiki/display/PLAT/How+to+Delete+a+Course This link that you shared, can this procedure be used to delete a course from stage?

PR #119, opened 2017-01-19
Do not fail tests if the saving notification is missed.

jzoldak:
This should solve the current flakiness for TE-1944. I'd like to get this in now. 

Out of scope for this PR but opportunities for later improvement/refactoring:
* Find the other usages in this repo and use it there too
* Submit as a new method upstream to the bok-choy repo. For this I would suggest keeping the wait_for_notification method name, and adding to it a keyword argument that will retain the default behavior and allow overriding to catch/swallow the exceptions.
* I'm not sure we even ever need to wait for the notification as a synchronization point anyhow. For this new logic, the worst case will be that we waste an extra 60 seconds (the default value in the bok-choy code) waiting for a promise that is broken. That seems however to be much less than the time we will spend troubleshooting future possible flakiness if we don't wait and the change is not yet saved.

FWIW I was able to run these changes locally, with it happening to both traverse through the new logic (it didn't see the popup) as well as the original (the notification was seen) logic.

### DIFF
--- a/regression/pages/studio/pages_page_studio.py
+++ b/regression/pages/studio/pages_page_studio.py
@@ -2,13 +2,13 @@
 Extended Pages page for a course.
 """
 from selenium.webdriver import ActionChains
-from edxapp_acceptance.pages.common.utils import (
-    wait_for_notification,
-    click_css
-)
+from edxapp_acceptance.pages.common.utils import click_css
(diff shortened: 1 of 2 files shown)


raeeschachar:
:+1: Fulfills our need

Muddasser:
👍

PR #120, opened 2017-01-20
Avoid UI login

Muddasser:
This PR is to address the issue stated in [TE-1939](https://openedx.atlassian.net/browse/TE-1939).

@jzoldak Please take a look. I have changed the code to avoid UI login for the LMS tests. For the studio test the process will be the same but the cookies would be different.

### DIFF
--- a/regression/tests/helpers.py
+++ b/regression/tests/helpers.py
@@ -9,8 +9,10 @@ from regression.pages.studio import BASE_URL
 from regression.pages import (
     BASIC_AUTH_USERNAME, BASIC_AUTH_PASSWORD, LOGIN_EMAIL, LOGIN_PASSWORD
 )
-from regression.pages.lms import LOGIN_BASE_URL
-
+from regression.pages.lms import LMS_BASE_URL
+from regression.pages.lms import LOGIN_BASE_URL as LMS_AUTH_URL
+from regression.pages.studio import STUDIO_BASE_URL
+from regression.pages.studio import LOGIN_BASE_URL as STUDIO_AUTH_URL
 
 COURSE_ORG = 'COURSE_ORG'
 COURSE_NUMBER = 'COURSE_NUMBER'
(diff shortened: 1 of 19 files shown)


regression/tests/helpers.py
```diff
@@ -90,3 +95,100 @@ def login(login_page):
...
         login_password = LoginHelper.get_login_password()
         login_page.login(login_email, login_password)
+
+
+class LoginApi(object):
+    """
+    An Api to login the stage.
+    """
+    def __init__(self):
+        self.login_url = 'https://courses.stage.edx.org/login'
```
jzoldak: Let's make this parameterized so that we can point to different environments.

regression/tests/helpers.py
```diff
@@ -90,3 +95,100 @@ def login(login_page):
...
+        self.session.headers = self.post_headers(
+            response.cookies['csrftoken']
+        )
+
+    def login(self):
+        """
+        Login to the stage.
+        """
+        self.create_base_session()
+        login_post_url = 'https://courses.stage.edx.org/' \
```
jzoldak: Same here - make this parameterized so that we can point to different environments.

regression/tests/helpers.py
```diff
@@ -90,3 +95,100 @@ def login(login_page):
...
+        Arguments:
+        browser: Browser to pass the session to.
+        """
+        self.login()
+        # To make cookies effective, we have to set the
+        # domain of the browser the same as that of the
+        # cookies. To do this, just visit a page of the
+        # same domain.
+        # Cookies require the domain to be ".stage.edx.org"
+        browser.get(LOGIN_BASE_URL + '/dashboard')
```
jzoldak: Leave an additional comment here that this will cause the browser to navigate to the login screen, but its form does not need to be filled out and submitted. Navigation to the next page will result in the browser being logged in.

Muddasser:
@jzoldak @benpatterson @raeeschachar Please review.

jzoldak:
@Muddasser couple more suggestions.

1.  Looks like StudioBaseTestClass in studio_base_test.py is still using LoginHelper in its cleanup method.
* I believe that you should be able to also replace that reference.
* And then I think you can remove the entire LoginHelper class.
* This change should be made before merging in this PR.

2. Make the BASE LMS and Studio URLs user-settable by OS environment variable.
* Acceptance criteria: You should be able to point these tests at a sandbox with the correct courses installed on it, and the tests give the same results there as they do on staging (barring any other deployed code or configuration differences.)
* Your call as to whether to do that in a follow-up PR. I think that might be the best option, as we will have this UI-less login working against staging in the meantime. And also to minimize the time spent dealing with merges, as this PR needed to modify all the testcase files.

raeeschachar:
👍

PR #124, opened 2017-01-25
include tests in common folder to be run

Muddasser:


Muddasser:
Solution to [LT-52](https://openedx.atlassian.net/browse/LT-52).

We need to make 'common' folder a python package so that tests present in it can be run. Double checked, pep8 and pylint are not missing these tests.

PR #125, opened 2017-01-25
Rewrite assets tests as a flow and refactor for speed

jzoldak:
FYI @estute @michaelyoungstrom 

**Note:** Moving forward we use the sync_on_notification() method from regression/pages/studio/utils.py in any extended page objects that do something that causes a notification to appear then disappear.


jzoldak:
@raeeschachar / @Muddasser Please see the new commit, which should eliminate the flakiness.

This needs a rebase after the login changes. Couldn't get to that today, but plan to first thing tomorrow.

PR #127, opened 2017-01-27
Remove HTML component test

jzoldak:
This particular test can and should live in the edx-platform repo itself,
as it it testing functionality contained solely in that repo.


jzoldak:
My misunderstanding. We do not want to remove the entire testcase file, just the one testcase class.

PR #129, opened 2017-01-31
Accept different environments

Muddasser:


Muddasser:
@benpatterson @jzoldak @raeeschachar Please review. This should enable us to run tests on sandbox.

jzoldak:
I tried running the test file lms/test_lms_login.py and test_remember_me fails because it doesn't see a cookie named "_gali" 
I put a breakpoint there and ran it in debug mode and indeed it is not there. Instead there are only two cookies: sessionid, and csrftoken.

* Should we expect this test case to pass on a sandbox?
* Do you know if the other testcases all pass? We should confirm that they do before calling this work done.

jzoldak:
@Muddasser I see now your email comments about the tests that you know fail and why.

> * lms/test_dashboard.py:DashboardTest.test_resume_course presumes that there is at least on section is present but it is not present when we import the course. This limitation should be avoided.
> * lms/test_instructor_dashboard.py:AnalyticsTest.test_analytics_link fails because it can't find the link.
> * studio/test_studio_home.py:StudioFooterTest.test_terms_of_service_link fails because it can't find the link.
> * studio/test_studio_home.py:StudioFooterTest.test_privacy_policy_link fails because it can't find the link.
> * studio/test_donations.py:DonationsTest.test_user_donations runs on the demo course from stage but we authenticate on the sandbox. The cause of failure is this mismatch in authentication.
> * lms/test_lms_login.LoginTest.test_remember_me checks a certain cookie in the browser but somehow it is failing on sandbox. Could not investigate further. 

I do think that we need to sort these out before merging / calling this work done.
I would suggest you add in branching logic, or if the whole testcase doesn't make sense against a sandbox (e.g. the footer links), then maybe a [@skipIf](https://docs.python.org/2/library/unittest.html#unittest.skipIf) decorator based on the URL would be a good choice.

benpatterson:
Could you add follow-up ticket(s) for the tests that had to be skipped? Put the ticket number in the code as comments for the skipIf condition. Then you are good to merge from my perspective.

Muddasser:
@estute Please have a look.

estute:
:+1: Good to go.

Muddasser:
@benpatterson Please approve the changes.

PR #131, opened 2017-02-03
add self to authors

estute:


estute:
jenkins run e2e

jzoldak:
@estute AUTHORS was a remnant from a former process. I'm thinking we should remove this file rather than keep it updated. What do you think?

PR #133, opened 2017-02-13
Test cancel upload fix

raeeschachar:


raeeschachar:
This should fix the flakiness. @benpatterson @jzoldak @Muddasser 
Please review.

PR #134, opened 2017-02-15
Kashif/ecommerce pages

kashifch:
@jzoldak, @benpatterson, @Muddasser, @raeeschachar 
Here is a new chunk for white label tests, these pages are related to Ecommerce. Once these are merged I will push the LMS pages too


benpatterson:
@kashifch how do these compare with the [ecommerce end-to-end tests](https://github.com/edx/ecommerce/tree/master/e2e)?

kashifch:
@benpatterson,

I haven't looked at these tests but I guess the page objects would be similar even if tests are different

PR #137, opened 2017-02-17
remove drag and drop test

jzoldak:
@raeeschachar @Muddasser @estute


jzoldak:
Will be addressed in another PR.

PR #142, opened 2017-02-27
update vagrant box to 16.04 w x server

estute:
@benpatterson @Muddasser @raeeschachar 
This installs the version of FF that we are using for Jenkins testing and Ficus devstack testing. It also gets x11 forwarding working.

In order to get dependency issues with Firefox 45 fixed, I needed to bump the Vagrant box image up to Ubuntu 16.04 (xenial). This is good, since Precise will no longer be officially supported in April. Unfortunately, Canonical is has created a broken vagrant image and shows no signs of wanting to fix it. Instead Vagrant recommends using a Xenial bento box, which is basically a packer template for creating vagrant images from a base image (see https://github.com/chef/bento/blob/master/ubuntu-16.04-amd64.json).

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -9,10 +9,15 @@ $script = <<SCRIPT
 
     # Install system requirements
     apt-get update -y
-    apt-get install -y git python2.7 python-pip python2.7-dev firefox dbus-x11 vim libjpeg-dev libxml2-dev libxslt1-dev
-    apt-get install -y libtiff4-dev libjpeg8-dev zlib1g-dev libfreetype6-dev liblcms2-dev libwebp-dev tcl8.5-dev tk8.5-dev python-tk
+    apt-get install -y xorg git python2.7 python-pip python2.7-dev dbus-x11 vim libjpeg-dev libxml2-dev libxslt1-dev
+    apt-get install -y libtiff5-dev libjpeg8-dev zlib1g-dev libfreetype6-dev liblcms2-dev libwebp-dev tcl8.5-dev tk8.5-dev python-tk
+    # Install the same version of Firefox installed on https://build.testeng.edx.org/
+    # for testing parity
+    apt-get install -y firefox=45.0.2+build1-0ubuntu1
+
     pip install virtualenv==1.10.1
     pip install virtualenvwrapper
+
 
     # Create a login script to create the virtualenv
     cat > /home/vagrant/.bash_profile <<INIT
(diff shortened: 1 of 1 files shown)


estute:
I changed the base box file to mirror what is used by the ficus devstacks (see https://github.com/edx/configuration/blob/cfabf00c0c2359b92f2121b1edc905f56c699f80/vagrant/base/devstack/Vagrantfile#L35)

raeeschachar:
:+1: Seems good

Vagrantfile
```diff
@@ -9,11 +9,16 @@ $script = <<SCRIPT
 
     # Install system requirements
```
benpatterson: I realize this is an old Vagrantfile...but we should be using an ansible play(s). Having the set of apt packages in here means a separate maintenance cycle if any other requirements arrive. 

That said, I think this is ok for now. Could you put a ticket in the backlog for cleanup later? (Or complete removal of the vagrantfile if we go the docker route.)

Vagrantfile
```diff
@@ -9,11 +9,16 @@ $script = <<SCRIPT
...
     apt-get update -y
-    apt-get install -y git python2.7 python-pip python2.7-dev firefox dbus-x11 vim libjpeg-dev libxml2-dev libxslt1-dev
-    apt-get install -y libtiff4-dev libjpeg8-dev zlib1g-dev libfreetype6-dev liblcms2-dev libwebp-dev tcl8.5-dev tk8.5-dev python-tk
+    apt-get install -y xorg git python2.7 python-pip python2.7-dev dbus-x11 vim libjpeg-dev libxml2-dev libxslt1-dev
```
benpatterson: We also use a different version of git from a ppa. 

We should be sure that we can compile pillow with jpg support on this box.

Other than that, let's just add this info to the ticket. I don't want to spend cycles perfecting until it's clear we'll be using it long-term.

EDIT: I see libjpeg8, which means pillow will compile fine. No need to check, but leaving the comment here just to pour more bits of data on you that may come in handy later. Pillow will successfully compile when there are system libs missing; it'll just not-support that area of its functionality. So, it can get tricky on you.

Vagrantfile
```diff
@@ -9,11 +9,16 @@ $script = <<SCRIPT
...
+    apt-get install -y libtiff5-dev libjpeg8-dev zlib1g-dev libfreetype6-dev liblcms2-dev libwebp-dev tcl8.5-dev tk8.5-dev python-tk
```
benpatterson: Ah there's libjpeg8. Does that mean libjpeg-dev is not necessary?

Vagrantfile
estute: I agree 100%. Totally a quick'n'dirty fix. Filed: https://openedx.atlassian.net/browse/TE-2013

PR #147, opened 2017-03-03
Fixed a bug in paver

Muddasser:


Muddasser:
@benpatterson @raeeschachar Please review. Paver was not running tests with flaky when we don't pin-point a test. i.e. paver e2e_test --with-flaky was failing on the grounds that --with-flaky was mistakenly taken as the name of file to run tests from.

pavelib/paver_utils.py
```diff
@@ -37,6 +36,9 @@ def command(report_name="report.xml", test_name=""):
             "--with-xunit",
             "--xunit-file='{}'".format(report_path)
             ]
+
+        for arg in arguments['cmd_args']:
+            construct_command.append(arg)
```
raeeschachar: extend instead of append? makes sense?

Muddasser: Oh yeah, good suggestion.

PR #148, opened 2017-03-04
Clean up install noise. Pipe to files.

benpatterson:


benpatterson:
@edx/testeng @cpennington This pipes various setup output into files and makes the console output much easier to read. PTAL.

PR #150, opened 2017-03-09
Update course test

Muddasser:


Muddasser:
@jzoldak @benpatterson @raeeschachar Please review guys. This is against this ticket.  https://openedx.atlassian.net/browse/LT-55. 

Is this what you were talking about?

benpatterson:
CC @jibsheet Can you recommend a way the Muddasser can verify that a celery task was involved with this transaction? For example, could he find it on splunk (and if so, could you suggest a query with the index for staging, etc)

jibsheet:
Sure, you could do it with splunk, there's an API, the query would looks something like
```index=stage-edx task source="/edx/var/log/supervisor/cms*" import```
but you'll need to differentiate between import_olx and course_import, that query is just to get your started.  I'm also not sure how you would know what task id you were assigned (since someone else might manually run an import during an e2e run).  The logs may or may not contain enough course identifying information.

@jmbowman might have other ideas for tracking this

Is the goal to check that the import is complete or confirm that it used celery?

Please note that I'm offline for the rest of the week so won't be following up on this until the 20th.

**Comment by benpatterson** [comment]
The goal was to confirm celery had been used. @Muddasser I'm going to assume that @jibsheet gave you enough info....you could run your tests, then run some queries to ensure that your particular transaction/user did the import and spawned celery work.

**Comment by Muddasser** [comment]
@benpatterson @jzoldak @raeeschachar A reminder for a review.

File: regression/pages/studio/import_course_page.py [hunk 1]
```diff
@@ -0,0 +1,36 @@
...
+        self.browser.execute_script(
+            '$(".file-name-block").show();$(".file-input").show()'
+        )
+        # Upload the file.
+        self.q(css='input[type="file"]')[0].send_keys(asset_file_path)
+        # Upload the same file again. Reason behind this is to decrease the
+        # probability or fraction of times the failure occur. Please be
+        # noted this doesn't eradicate the root cause of the error, it
+        # just decreases to failure rate to minimal.
+        # Jira ticket reference: TNL-4191.
```
benpatterson: Do we still need to run this twice? I see TNL-4191 has been resolved. Perhaps the same approach needs to be taken here (waiting for the click handler)?

PR #151, opened 2017-03-16
Another flaky test on master

raeeschachar:


raeeschachar:
@edx/testeng 
While testing of https://openedx.atlassian.net/browse/TE-1941, this was another test that was flaky on master. Added flaky decorator to it for now, so the release pipeline isn't effected.

PR #149, opened 2017-03-07
Refactoring tests

raeeschachar:

PR #154, opened 2017-03-21
Fix flaky logout on FF 45

raeeschachar:


raeeschachar:
@edx/testeng 
Logout test flakiness was fixed after [refactoring tests PR](https://github.com/edx/edx-e2e-tests/pull/149). We are good to remove this from flaky ticket.

PR #155, opened 2017-03-22
Grade range fix on ff 45

raeeschachar:


raeeschachar:
@edx/testeng 
These were the tests failing on Master yesterday. I was able to fix `test_grade_range` (still have to test it on master). I have marked `test_hide_and_show_pages` flaky again and reopened https://openedx.atlassian.net/browse/LT-66 as I am still unsure why it always passes locally and on jenkins pr job and fails on master. 
After this merge I will be watching release pipeline runs to make sure they are not blocked.

PR #159, opened 2017-03-28
Raees/test misc

raeeschachar:


raeeschachar:
@edx/testeng 
Please review.

regression/pages/whitelabel/home_page.py
```diff
@@ -36,3 +36,68 @@ def click_login_button(self):
...
+        """
+        Get list of social media links
+        Returns:
+            social media links:
+        """
+        footer_css = '.footer-main a'
+        social_links = []
+        face_book = self.q(
+            css=footer_css
+        ).filter(lambda elem: elem.text == 'Facebook')
```
jzoldak: nit: No need to fix for this PR. Each of these .q().filter() commands will query the DOM element tree and pick out the element that matches. There is sure to be a way to grab the footer once and then process it.

PR #160, opened 2017-03-30
Added flaky to unit studio view test

raeeschachar:


raeeschachar:
@edx/testeng cc: @cpennington 
Another test that is flaky. I have marked it flaky for now and restructured a bit for cleanup with in test. I think because of platform page object dependencies, we will have to recheck these tests once a while. Please review.

PR #162, opened 2017-04-03
Fix flakiness of show hide pages test

raeeschachar:


raeeschachar:
@edx/testeng 
I am optimistic that this would fix the flakiness. Please review.

regression/tests/studio/test_course_pages.py
```diff
@@ -123,22 +121,29 @@ def setUp(self):
...
         """
         Verifies that user can View live course from pages page
         """
         self.pages_page.click_view_live_button()
-        courseware_page = CoursewarePage(
-            self.browser, get_course_info())
-        courseware_page.wait_for_page()
+        self.courseware_page.wait_for_page()
 
     def test_drag_and_drop_of_pages(self):
```
jzoldak: Please remove this drag and drop test from the test suite? My reasoning:
* The functionality it is testing (drag and drop on the courseware page in studio) is already covered in the bok-choy test suite within the edx-platform itself. 
* Having it in the end to end tests does not give us any more information about the quality of the codebase, or about the configuration/integration of the parts of the system in the staging environment.
* And its downside of being here is that it is likely to be flaky
* as well as increasing the execution time of the test suite.

regression/tests/studio/test_course_pages.py
```diff
@@ -123,22 +121,29 @@ def setUp(self):
...
             self.browser,
             self.course_info['org'],
             self.course_info['number'],
             self.course_info['run']
         )
+        # Login to LMS to avoid authentication issue
+        self.course_page.visit()
         self.pages_page.visit()
 
     def test_view_live_pages(self):
```
jzoldak: Please remove this testcase.
If the functionality that it tests is broken, then we will be alerted by the assert_page_is_shown_in_lms test which starts off by doing the same thing.

raeeschachar:
@jzoldak 
Done. Please review.

PR #181, opened 2017-05-11
Tests for otto existing user

raeeschachar:


raeeschachar:
jenkins run microsites

raeeschachar:
@benpatterson @jzoldak @Muddasser @estute 
Please review.

raeeschachar:
FYI, some tests were removed as they were being covered somewhere else and we didn't need them here, as per @kashifch 
So we decided the ones needed and added in this PR.

regression/tests/whitelabel/test_otto_existing_user.py
```diff
@@ -0,0 +1,79 @@
...
+        self.pay_with_cybersource()
+        self.dashboard_page.wait_for_page()
+        self.assert_enrollment_and_logout()
+
+    def test_select_course_and_login(self):
+        """
+        Scenario: Otto flow - A registered user is able to select a course,
+        login  and make payment for the course using the credit card
+        """
+        self.addCleanup(
```
jzoldak: nit: looks like this could go in the setUp method.

raeeschachar:
@edx/testeng 
Need one more approval.

raeeschachar:
jenkins run microsites

raeeschachar:
jenkins run microsites

raeeschachar:
jenkins run microsites

raeeschachar:
jenkins run microsites

raeeschachar:
jenkins run microsites

raeeschachar:
jenkins run microsites

raeeschachar:
jenkins run microsites

regression/pages/whitelabel/home_page.py
```diff
@@ -107,3 +107,9 @@ def go_to_registration_page(self):
         Go to registration page
         """
         self.q(css='.btn-brand.btn-client[href="/register"]').click()
+
+    def go_to_courses_page(self):
+        """
+        Go to courses page
+        """
+        self.q(css='.brand-link[href="/courses"]').click()
```
jzoldak: should this have a wait_for_page after it?

regression/tests/whitelabel/course_enrollment_test.py
```diff
@@ -201,3 +202,31 @@ def assert_enrollment_and_unenroll(self):
...
+        self.receipt_page.click_in_nav_to_go_to_dashboard()
+
+    def verify_receipt_info(self):
+        """
+        Verify that Course title, Course Price, total price and order date on
+        receipt page is correct.
+        """
+        self.assertIn(self.course_title, self.receipt_page.order_desc)
+        self.assertEqual(
+            unicode(datetime.datetime.utcnow().date()),
```
jzoldak: note that there is a slight chance that this will fail if the test execution crosses a boundary of midnight. That said, I think that the risk is so slight that it's fine to leave it like this. May be worth a comment in the code though.

regression/pages/whitelabel/home_page.py
raeeschachar: Waiting for the page in the [test](https://github.com/edx/edx-e2e-tests/pull/181/files#diff-1ace6cae9636a71d17f88c7d3df4641dR57) actually. Does that make sense?

regression/tests/whitelabel/course_enrollment_test.py
raeeschachar: Done

regression/pages/whitelabel/home_page.py
jzoldak: My concern is that there might be other tests written in the future that use this method, and they could be flaky if the author doesn't know or remember to put a wait_for_page in **their** test. 
I'm now realizing that this action changes what page you're on. 
So maybe the best way to mitigate that risk is to add a comment here reminding people that if they use this method, then they need a wait_for_page statement in their test right after it.

raeeschachar: Ok, right. Adding.

raeeschachar:
@jzoldak @Muddasser @estute 
Please review.

raeeschachar:
jenkins run microsites

PR #184, opened 2017-05-16
test_dynamic_enrollment_coupon

Muddasser:


Muddasser:
@raeeschachar These are due to changes in the UI of the whitelabel sites. I have changed some of the instances to ensure that my tests start working. I hope once your PR(For fixing these failing tests formally) is merged, these tests would get fixed automatically.

Muddasser:
@raeeschachar I believe you are already working on the test related to donations which is failing on edX stage.

regression/tests/whitelabel/test_dynamic_enrollment_coupon.py
```diff
@@ -0,0 +1,134 @@
...
+            course_seat_types=COURSE_SEAT_TYPES['prof'],
+            stock_record_ids=[],
+            max_uses=2
+        )
+        self.coupon.setup_coupons_using_api(self.course_price)
+        coupon_code = self.coupon.coupon_codes[0]
+        # Login to application using the existing credentials
+        coupon_users = list(COUPON_USERS.values())
+        last_user = len(coupon_users) - 1
+        for i, coupon_user in enumerate(coupon_users):
```
jzoldak: nit: a way to make this code more readable would be to loop through all but the last user (using list slice notation), and then pull the part that you have out of the else clause and do that separately. Something like:
for coupon_user in coupon_users[:-1]:
    ...
redeem_coupon = RedeemCouponPage...

PR #185, opened 2017-05-17
Fix for failing user donation test

raeeschachar:


raeeschachar:
@Muddasser @jzoldak @benpatterson @estute 
Please review.

PR #189, opened 2017-06-06
Skip tests broken by https://github.com/edx/ecommerce/pull/1252

cpennington:


jibsheet:
@cpennington I can't tell if we didn't skip hard enough from the test failures.

I'm :+1: on this once tests pass.  Some decision on these tests needs to be reached between @douglashall  or whoever he nominates from WhiteLabel and @schenedx / @mjfrey  (I think the ecom behavior is the correct behavior, but I also thing the tests are testing behavior we don't want to lose).

My problem is that these tests haven't passed for more than a week now and no team seems to have a ticket to fix them.  Skipping them so we can release with less confusion / digging seems the correct stopgap.

jzoldak:
@raeeschachar @aisha-sajid FYI

PR #192, opened 2017-06-09
e2e TE-2010

aisha-sajid:
Added environment variables for microsites, how and where to add new tests and links to the guide for setting up development environment

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -37,6 +37,7 @@ We recommend using the provided Vagrant environment to develop and run tests.
 
 You will also need a deployed installation of edX lms and studio to run the tests against.
 See `edx/configuration <http://github.com/edx/configuration>`_ for instructions on provisioning an edX instance.
+For additional information on how to setup your development environment, see `Developer Onboarding <https://openedx.atlassian.net/wiki/pages/viewpage.action?spaceKey=ENG&title=Developer+Onboarding#DeveloperOnboarding-Step4:Getreadytodevelop>`_
 
 
 
@@ -72,13 +73,8 @@ Configuration
     paver install_pages
 
 
-How to run tests?
----------------------
-Make sure you are in the correct folder.
-
-.. code:: bash
-
-    cd $HOME/edx-e2e-tests
+How to run LMS and Studio tests
+--------------------------------
 
 Before running tests, please ensure that following environment variables are set.
 
@@ -88,13 +84,16 @@ Before running tests, please ensure that following environment variables are set
     ==> BASIC_AUTH_PASSWORD
     ==> USER_LOGIN_EMAIL
     ==> USER_LOGIN_PASSWORD
-
+    ==> COURSE_RUN
+    ==> COURSE_DISPLAY_NAME
+    ==> COURSE_NUMBER
+    ==> COURSE_ORG
 
 To run all the tests:
 
 .. code:: bash
 
-    paver e2e_test
+    paver e2e_test --exclude=whitelabel
 
 
 The commands also accept nose-style specifiers for test case or module:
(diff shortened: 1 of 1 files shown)


estute:
jenkins run microsites

raeeschachar:
@aisha-sajid 
You might need to rebase off master before merging.

README.rst
```diff
@@ -124,9 +128,69 @@ To update page objects installed from external repos:
...
+    ==> ACCESS_TOKEN
+
+
+To run all the tests in the file:
+
+.. code:: bash
+
+    paver e2e_test whitelabel/test_user_account.py
+
+To run all the tests in a particular class:
```
estute: I think we can remove the running specific classes/tests parts from the white label section (since it is the same as the e2e section, except for the paver target. Instead, you could just make a note about using the same Nose test specifiers as mentioned above, adding that you must use the white label paver target. What do you think?

README.rst
```diff
@@ -88,7 +89,10 @@ Before running tests, please ensure that following environment variables are set
     ==> BASIC_AUTH_PASSWORD
     ==> USER_LOGIN_EMAIL
     ==> USER_LOGIN_PASSWORD
-
+    ==> COURSE_RUN
+    ==> COURSE_DISPLAY_NAME
+    ==> COURSE_NUMBER
+    ==> COURSE_ORG
 
 To run all the tests:
```
estute: On second thought... @raeeschachar, maybe the paver e2e_test target should exclude the white label tests by default, and only run them with an include flag? Either way, I think we should try to be a little more clear about what this paver target does, because it doesn't exactly run all of the tests.

raeeschachar: @estute 
by default we are excluding white label tests on Jenkins:
https://github.com/edx/edx-e2e-tests/blob/master/jenkins/end_to_end_tests.sh#L32

I think we can specify:
- Run all tests 'paver e2e_test'
- With flaky 'paver e2e_test --with-flaky'
- Exclude white label 'paver e2e_test --exclude=whitelabel'
- With flaky without whitelabel 'paver e2e_test --with-flaky --exclude=whitelabel'
Does that make sense?

estute: @raeeschachar Yes. I meant that 'paver e2e_test' should exclude the whitelabel tests without using the '--exclude' flag. I am thinking the case where you want to run both normal and whitelable tests in one command is very rare, but I may be wrong. Thoughts?

raeeschachar: @estute ah understood. We had in mind the following:
paver e2e_test (all tests)
paver e2e_test studio (tests in studio folder)
paver e2e_test whitelabel (tests in whitelabel folder)
paver e2e_test lms (tests in lms folder) 
So anyone running tests has all the options. But your idea is correct too, it's a rare chance that someone might need to run all the tests. I think we can do this too i.e --exclude-wl by default.
We can choose any of the two.

README.rst
```diff
@@ -72,8 +73,8 @@ Configuration
     paver install_pages
 
 
-How to run tests?
----------------------
+How to run LMS and Studio tests?
```
aisha-sajid: (Y)

aisha-sajid: done

README.rst
```diff
@@ -124,9 +128,69 @@ To update page objects installed from external repos:
 
     paver install_pages
 
+How to run Whitelabel tests?
+----------------------------
+
+Make sure you are in the correct folder.
```
aisha-sajid: done

README.rst
aisha-sajid: yeah I guess it's repetitive. I'll change it accordingly

aisha-sajid: done

README.rst
```diff
@@ -124,9 +128,69 @@ To update page objects installed from external repos:
...
+    ==> TEST_EMAIL_ACCOUNT
+    ==> TEST_EMAIL_PASSWORD
+    ==> ACCESS_TOKEN
+
+
+To run all the tests in the file:
+
+.. code:: bash
+
+    paver e2e_test whitelabel/test_user_account.py
```
aisha-sajid: done

README.rst
aisha-sajid: @raeeschachar  I think we should include the flaky tests and exclude flag in a note after the "How to run tests" section. 

However, I'm not sure whether the e2e_test flag exludes whitelabel tests by default or not. Need to check that up. 

as per the above discussion with @estute I can't figure out what have we agreed upon as our final choice out of the two?

Also, in another conversation on this same PR, @estute  has mentioned that we use a different paver target for running the whitelabel tests i.e **e2e_wl_test**. so in that case I guess **e2e_test** paver target will not run whitelabel tests. Thoughts?

estute: @aisha-sajid concerning the flaky tests, I would not put it into the documentation. We do not do it for the edx-platform. Plus, we are in talks concerning removing the flaky test library in the near future.

estute: As for the other question: For now, I think the documentation should reflect exactly how we run tests on Jenkins. Once we have our paver targets in order AND update the jenkins scripts, we can update the README to reflect this. For the time being, I think we should just have:
* How to run all the e2e tests -> paver e2e_test --exclude=whitelabel
* How to run a specific e2e test -> paver e2e_test (+ nose specifier)
* How to run wl/microsites tests -> paver e2e_wl_test (+ nose specifiers)

The rationale for splitting them for the time being, is that they do not consume the same environment variables. We want people who have never used this repo to be able to easily test/run code the way it will be run on our CI systems. @raeeschachar  does this make sense to you?

raeeschachar: @estute @aisha-sajid 
Yup I think it makes sense.

aisha-sajid: @raeeschachar  @estute 
I have made the specified changes. Please review

PR #194, opened 2017-06-15
e2e-TE1914 

aisha-sajid:
Added Dockerfile


estute:
@aisha-sajid  could you rebase this off of master instead of e2e-TE2010, as that branch is still in progress and subject to change?

PR #116, opened 2017-01-10
[WIP] Update readme with instructions for using the browser from a container

jzoldak:

PR #195, opened 2017-06-20
e2e-TE1914 - Dockerfile

aisha-sajid:
Added dockerfile in e2e-tests repo

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,14 +3,20 @@ FROM python:2.7-onbuild
 
 USER root
 
+#Add the e2e-repo to the container
+ADD https://github.com/edx/edx-e2e-tests.git /edx_e2e_tests
+
 #Configuration
 RUN apt-get update
-RUN apt-get install git
+RUN apt-get install python-setuptools python-dev build-essential
+RUN easy_install pip
 RUN pip install paver
-RUN pip install bok_choy
 
 RUN pip install virtualenv==1.10.1
 RUN pip install virtualenvwrapper
+
+RUN apt-get install git
+RUN pip install bok_choy
 
 RUN pip install -r requirements/base.txt
 RUN paver install_pages
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,13 @@
-#Using the official ubuntu OS as a Base Image
-FROM ubuntu:16.04
+#Using the official python 2.7 as a Base Image
+FROM python:2.7-onbuild
 
 USER root
-
-#Add the e2e-repo to the container
-ADD https://github.com/edx/edx-e2e-tests.git /edx_e2e_tests
 
 #Configuration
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+FROM ubuntu:16.04
+
+USER root
+
+#Configuration
+RUN pip install -r requirements/base.txt
+RUN paver install_pages
+
+#Setting up environment variables. Change the value according to the value of the ENV_VAR
+
```
estute: We need a few more environment variables here. You might want to make use of local_env.sh

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+ENV BASIC_AUTH_USER value	
+ENV BASIC_AUTH_PASSWORD value
+ENV USER_LOGIN_EMAIL value
+ENV USER_LOGIN_PASSWORD value
+
+#Edit the CMD command to run speciifc tests
```
estute: If we use the CMD directive to run the tests, it will require a user to rebuild the image every time they want to run a different paver command, i.e running a specific test/test suite. I think it would be better to have the ENTRYPOINT/CMD just be something like /bin/bash. You could run the container with the '-t' and '-d' flags to have it run detached, and then shell into the container to run arbitrary paver commands

Dockerfile
```diff
@@ -0,0 +1,18 @@
+#Using the official ubuntu OS as a Base Image
+FROM ubuntu:16.04
+
+USER root
+
+#Configuration
```
estute: These files will need to present in order to build the image. For the pip install, you could just do an ADD/COPY. However, I tend to think volumes might be better for this. I suggest two volumes- one for the e2e test infrastructure and one for the edx-platform repo. That way, you can easily symlink them to edx-e2e-tests/lib and run install_pages.

Dockerfile
```diff
@@ -0,0 +1,18 @@
+#Using the official ubuntu OS as a Base Image
+FROM ubuntu:16.04
+
+USER root
+
```
estute: There are additional libraries that are probably necessary for the system. Check out https://github.com/edx/edx-e2e-tests/blob/master/Vagrantfile

Dockerfile
estute: ^ adding on that: it would be best if the user could shell directly into the directory containing the tests AND have the virtual environment activated (if we continue to use it)

estute:
@aisha-sajid Take a look at https://github.com/edx/edx-e2e-tests/pull/116. JZ has stubbed out some documentation for using a Docker container with Firefox for running the bok-choy tests. We would like to end up with two containers- a container with the tests/page objects that drives another container with a browser. We should use Docker-compose for this. In any case, can you please include some documentation on the uses of the container(s).

PR #198, opened 2017-07-11
Fix Studio link to video documentation

srpearce:
Updated URL in regression/tests/studio/test_video_upload_help.py to reflect new location of video documentation.


jibsheet:
@jzoldak we're seeing some failures on the tests, but they're not related to Sylvia's change (which is a change to a test with a hardcoded URL) so we're going to go ahead and merge.

PR #197, opened 2017-07-07
Remove flaky requirement from e2e tests

raeeschachar:

PR #199, opened 2017-07-12
Removing test because it is flaky

raeeschachar:


raeeschachar:
@estute @jzoldak @aisha-sajid 
Please review. 
This is the first part of https://openedx.atlassian.net/browse/TE-2144, second part is https://github.com/edx/edx-e2e-tests/pull/197

PR #201, opened 2017-07-17
WIP - dockerize e2e development environment

raeeschachar:


README.rst
```diff
@@ -39,7 +39,18 @@ You will also need a deployed installation of edX lms and studio to run the test
...
+
+1. git clone https://github.com/edx/edx-e2e-tests
+
+2. docker-compose up --build
+
+3. docker exec -it edxe2etests_e2e-tests_1 /bin/bash
+
+4. source local_env.sh (make sure you have replaced the local_env.sh.sample)
 
+5. run lms and studio tests, as given belows
```
raeeschachar: nit: below

raeeschachar:
@jzoldak @estute 
Please review. The tests are working using docker containers. VNC support is still pending as there is not much support for that.

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN pip install virtualenv==1.10.1
+RUN pip install virtualenvwrapper
+
+RUN apt-get install git
+
+RUN pip install -r requirements/base.txt
+RUN paver install_pages
+
+# Set up environment variables
+CMD sleep 10000
```
estute: I understand using this for debugging, but we will need another solution for this. We need a CMD/ENTRYPOINT that will run indefinitely. The trouble is, all of the processes we want to run from within this container are short lived. At the least, you could use 'sleep infinity'. Perhaps there is a way to do this via the docker-compose file?

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+ADD . /edx-e2e-tests
+WORKDIR /edx-e2e-tests
+
+# Configuration
+RUN apt-get update
+RUN pip install paver
+
+RUN pip install virtualenv==1.10.1
+RUN pip install virtualenvwrapper
```
estute: We need to activate the virtualenvironment

Dockerfile
```diff
@@ -0,0 +1,20 @@
+# Using the official python 2.7 as a Base Image
+FROM python:2
+
+ADD . /edx-e2e-tests
+WORKDIR /edx-e2e-tests
+
+# Configuration
+RUN apt-get update
```
estute: A tip for reducing the size of Docker containers is to combine similar commands into a single RUN command. So, for example, you could combine both of the apt-get installs into one.

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN pip install paver
```
estute: With the exception of the virtualenv, I think all pip installs should be coming from requirements. This way, there can be no difference between the versions run by Jenkins and locally

README.rst
```diff
@@ -39,7 +39,18 @@ You will also need a deployed installation of edX lms and studio to run the test
...
 For additional information on how to setup your development environment, see `Developer Onboarding <https://openedx.atlassian.net/wiki/pages/viewpage.action?spaceKey=ENG&title=Developer+Onboarding#DeveloperOnboarding-Step4:Getreadytodevelop>`_
 
+OR Using Dockers
+------------
+
+1. git clone https://github.com/edx/edx-e2e-tests
+
+2. docker-compose up --build
+
+3. docker exec -it edxe2etests_e2e-tests_1 /bin/bash
```
estute: Where does the name of the container come from? I think you may need to specify it in the docker-compose file

raeeschachar:
@estute @jzoldak 
Done. Please review.

README.rst
```diff
@@ -39,7 +39,35 @@ You will also need a deployed installation of edX lms and studio to run the test
 See `edx/configuration <http://github.com/edx/configuration>`_ for instructions on provisioning an edX instance.
 For additional information on how to setup your development environment, see `Developer Onboarding <https://openedx.atlassian.net/wiki/pages/viewpage.action?spaceKey=ENG&title=Developer+Onboarding#DeveloperOnboarding-Step4:Getreadytodevelop>`_
 
+OR Using Dockers
+------------
+
+1. Clone the repo:
+
+.. code:: bash
+    git clone https://github.com/edx/edx-e2e-tests
```
jzoldak: The code snippets are not displaying in the rendered view.
I'm pretty sure that you need an empty line above them.

Dockerfile
```diff
@@ -0,0 +1,18 @@
+# Using the official python 2.7 as a Base Image
+FROM python:2
+
+ADD . /edx-e2e-tests
+WORKDIR /edx-e2e-tests
+
+# Configuration
+RUN apt-get update && apt-get install git
+RUN pip install virtualenv==1.10.1
+RUN pip install virtualenvwrapper
```
estute: We must create the venv before activating it

raeeschachar:
@estute @jzoldak 
Please review.

estute:
@raeeschachar I tried to run this, and it seems that the page objects were not installed correctly on the python container. Can you verify that they are when you run docker compose?

raeeschachar:
@estute @jzoldak 
I was seeing a connection refused issue, fixed it in this commit. Please try to use docker as per the readme file to see if anything needs to be added/removed. Please review once again. Thanks

PR #203, opened 2017-07-26
first login test

kashifch:
@raeeschachar @jzoldak @brittneyexline @douglashall 

Here is the first test for enterprise login. For the time being I have kept the enterprise tests separate cause I don't want these tests to break the e2e or microsite tests.

Later on depending on the volume of tests we cane decide whether to merge these tests with e2e or run these as a separate job

### DIFF
--- a/regression/pages/enterprise/enterprise_const.py
+++ b/regression/pages/enterprise/enterprise_const.py
@@ -1,0 +1,22 @@
+"""
+URLs and constants for enterprise stuff
+"""
+
+import os
+
+
+DEFAULT_IDP_URL = u"https://idp.testshib.org/idp/Authn/UserPassword"
+
+IDP_URL = os.environ.get('IDP_URL', DEFAULT_IDP_URL)
+
+DEFAULT_IDP_NAME = "TestShib"
+
+IDP_NAME = os.environ.get('IDP_NAME', DEFAULT_IDP_NAME)
+
+DEFAULT_ENTERPRISE_NAME = 'TestDemoEnterprise'
+
+ENTERPRISE_NAME = os.environ.get('ENTERPRISE_NAME', DEFAULT_ENTERPRISE_NAME)
+
+DEFAULT_IDP_CSS_ID = 'arbi-test-shib'
+
+IDP_CSS_ID = os.environ.get('IDP_CSS_ID', DEFAULT_IDP_CSS_ID)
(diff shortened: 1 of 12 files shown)


kashifch:
Included the 2nd test as well, since one test assumes that user will be linked with IDP and the other test assumes that user will not be linked, I have changed the flow a little bit and assumed linking as default state. Unlinking is done prior to 2nd test every time

kashifch:
@raeeschachar @jzoldak @brittneyexline @douglashall 
please review

regression/pages/enterprise/enterprise_const.py
```diff
@@ -0,0 +1,22 @@
...
+"""
+
+import os
+
+
+DEFAULT_IDP_URL = u"https://idp.testshib.org/idp/Authn/UserPassword"
+
+IDP_URL = os.environ.get('IDP_URL', DEFAULT_IDP_URL)
+
+DEFAULT_IDP_NAME = "TestShib"
```
brittneyexline: Is this all based on an existing enterprise customer and tpa provider configuration? where does that configuration live?

kashifch: At the moment I am using TestDemoEnterprise as default enterprise and testshib as default IDP, The values will be stored in environment variables so user can specify different values if required. In case no value is specified test will run with the default settings

kashifch:
Hi @brittneyexline,

Yes at the moment I just want the overall enterprise setup to get merged in the master so I can build further tests on top of it, I am making some other changes in the microsite tests as well so I don't want too many interdependent changes.  

These initial tests will eventually become helper functions to be used in complex flows

zubair-arbi:
@edx/testeng could someone please prioritize review for this PR as this is base work for further enterprise end to end tests.

regression/tests/helpers/api_clients.py
```diff
@@ -140,9 +140,11 @@ class LmsLoginApi(LoginApiBaseClass):
     """
     Login api for LMS
     """
-    def __init__(self):
+    def __init__(self, target_page=None):
```
zubair-arbi: nit: `target_page_name='/dashboard'`, this way you don't need to set this value again in method.

zubair-arbi:
@kashifch just a nit, otherwise LGTM 👍

PR #204, opened 2017-07-26
PLAT-1590 Support for running in Docker Devstack

jmbowman:
Initial support for running the tests under Docker Devstack.  This should make it easier to develop the tests (since you can run them locally and watch the browser via VNC) and sanity check our Docker image builds.  I also fixed issues in a few of the tests that were causing them to fail occasionally due to race conditions.

* Added optional support for non-HTTPS target URLs
* Added optional support for target URLs which don't have HTTP basic authentication enabled
* Allow testing courses specified via the environment variables documented in the README (a couple of tests were still assuming a specific hard-coded course key)
* Put some checks in asset deletion to make sure the page is in the expected state before clicking
* Made sure that the click handler for the "Add Page" button is registered before clicking on it
* Made sure that the page has time to update before each click on the way to add a custom JS component
* Upgrade bok-choy and its dependencies
* Put in a similar fix to [what we did in edx-platform](https://github.com/edx/edx-platform/blob/master/common/test/acceptance/tests/helpers.py#L740) to allow us to keep using the old version of Firefox on Jenkins with the newer version of Selenium we're upgrading to
* Allow specification of the expected CMS help base URL, since it's different for Docker Devstack
* Avoided a second attempt to click on Publish after adding a word cloud
* Skip one test under Docker Devstack which requires the video upload pipeline to be configured
* Updated the course tarball to use the edX organization and "E2E-101" course name to make it clearer what the course is for (via [this feedback](https://github.com/edx/configuration/pull/4020#discussion_r129950051) on a related PR).
* Made a few tweaks to the Dockerfile to work better for uploading to Docker Hub and using in Docker Devstack.
* Support having Travis automatically build and upload new Docker images on commits to master.
* Added an environment settings file for use in testing the services running in Docker Devstack.
* Switched the exposed VNC port for the Firefox container to one which won't conflict with macOS screen sharing or Docker Devstack's Chrome and (newer version of) Firefox containers.

With these changes, I have all the tests passing locally with the latest version of Chrome.  (There are some bugs in WebDriver support with recent versions of Firefox that cause a few of the tests to fail there.)


jmbowman:
It looks like the one failing test here has failed randomly before, for example in https://build.testeng.edx.org/view/e2e-tests/job/edx-e2e-tests/603/ .  Has any investigation into that already been done?

jmbowman:
@estute Any feedback on this, especially with the Dockerfile and docker-compose.yml updates?  And @clintonb , does this Dockerfile look better to you than the one which was in the configuration repo PR?  Does the Travis configuration for building and upload images look like it should work once the Docker Hub username and password secure environment variables are added?

Dockerfile
```diff
@@ -1,15 +1,27 @@
...
-
 # Install requirements and pages
-RUN pip install -r requirements/base.txt
-RUN paver install_pages
+# Deletes the edx-platform checkout afterwards, it will be mapped in from the host
+RUN pip install -r requirements/base.txt \
+    && paver install_pages \
+    && rm -rf /edx-e2e-tests/lib
 
 # Set up environment variables
```
clintonb: This comment does not apply to the line that follows the comment.

.travis.yml
```diff
@@ -11,3 +14,13 @@ script:
   - paver install_pages
```
clintonb: This will be done by the call to `docker-compose build` below. If you build the image first, you can run all of the validation inside a container and avoid duplicating work. We do something similar for credentials: https://github.com/edx/credentials/blob/master/.travis.yml.

jmbowman: Not a bad idea, but I'm worried the scope of this PR has already gotten big enough without fiddling around with how the tests get run in Travis.  I'm inclined to save that for a separate PR unless there's a reason it really has to be done at the same time as the rest of this.

jmbowman:
@jzoldak @estute - Any feedback on this, or should I merge it as is?

PR #205, opened 2017-07-31
Fix flakiness in creating new sections and subsections

jmbowman:
Click on the status row instead of the overall content area to finish editing a section or sub-section name; the old behavior sometimes inadvertently resulted in a click on the new unit button within that area.  This has caused several random test failures recently.


jmbowman:
@jibsheet Still waiting for tests to finish, but here's the PR.

jmbowman:
The tests passed, GitHub just lost track of the result due to an outage.  Merging.

PR #211, opened 2017-08-22
Fix target css for TOS page

jzoldak:
@estute @michaelyoungstrom 

The e2e tests in the CD pipeline are failing due to this css change.
I looked quickly to relate it back to a commit that changed the css class, but it wasn't immediately apparent so I am fixing and tagging you guys for a thumbs up. This should get us back to green.


jzoldak:
Merging with jenkins/microsites failing, as those tests are failing off current master.

PR #212, opened 2017-08-25
remove flaky test_pages_crud test

MichaelRoytman:
[EDUCATOR-1246](https://openedx.atlassian.net/browse/EDUCATOR-1246)

PR #213, opened 2017-08-29
Wait for js to be loaded on the Pages page  TE-2212

jzoldak:
@MichaelRoytman this should fix the underlying problem of #212.


MichaelRoytman:
@jzoldak Great! Thank you!

PR #215, opened 2017-08-31
registration and enrollment page

kashifch:
@brittneyexline, @zubair-arbi 

Tests for registration and enrollment page


zubair-arbi:
@edx/testeng Could someone of you please review these changes before we merge them?
cc @jzoldak

kashifch:
@brittneyexline We have a test for login via linked user, login via unlinked user, and one for registering new user but I think you are right that for course verification and stuff we are mostly using login, I will add new tests to do a mix and match of registration and login

PR #216, opened 2017-08-31
Kashif/registration form

kashifch:


kashifch:
@jzoldak, @estute 

There is a change in HarvardGlobalMed registration form which is breaking the microsite tests, I have fixed the issue, please review

PR #217, opened 2017-09-06
use api to register

kashifch:


kashifch:
@jzoldak @estute 

This is a fix for https://openedx.atlassian.net/browse/WL-1058. For now I have modified the Login API to support registration as well. The idea is to register user using api and then transfer the session to bok-choy just like login. This would reduce time significantly and allow me to unskip the wl tests. For now I have just use this API based registration for one test to make sure that it works alright. Once this PR is merged I will implement this for all tests.

Another change that I made is to separate the Logout API from Login API, Although it was convenient to use the login session to logout as well but I was thinking that it would be a constraint in scenarios where we would like to logout using API when the login was done through UI. So I have used the reverse approach for logout API and get cookies from current browser window and transferred these to API and use this session to logout user.

Please Review.

@bill-filler 
I was thinking if anyone from WL team could do one review as well, it would make it easier for testeng team, plus hopefully it would give the dev team a chance to understand what is going on in these tests and also perhaps they could identify logical issues more easily

bill-filler:
@kashifch I will take a look at this soon. Thank you

kashifch:
genkins run e2e

PR #220, opened 2017-09-19
get access token value for client

kashifch:


regression/pages/whitelabel/const.py
```diff
@@ -8,15 +8,16 @@
...
 
-ACCESS_TOKEN = os.environ['ACCESS_TOKEN']
+# Client information for access token
+OAUTH_CLIENT_ID = os.environ.get('OAUTH_CLIENT_ID')
+OAUTH_CLIENT_SECRET = os.environ.get('OAUTH_CLIENT_SECRET')
 
-# Select the Org for which to run the tests, Default is MITProfessionalX
-ORG = os.getenv('ORG', 'MITProfessionalX')
+# Select the Org for which to run the tests, Default is HarvardMedGlobalAcademy
+ORG = os.environ.get('ORG', 'HarvardMedGlobalAcademy')
```
zubair-arbi: Why we are setting `HarvardMedGlobalAcademy` as default organization? as previously it was `MITProfessionalX`. I think we can either change the placeholder name or there should be no default.

regression/pages/whitelabel/const.py
```diff
@@ -8,15 +8,16 @@
 AUTH_USERNAME = os.environ['BASIC_AUTH_USER']
 AUTH_PASSWORD = os.environ['BASIC_AUTH_PASSWORD']
 
-# Get API Access token
+# Global password
+PASSWORD = os.environ.get('GLOBAL_PASSWORD')
```
zubair-arbi: What is this password `GLOBAL_PASSWORD` for? If its basic auth then please rename the variable name.

regression/pages/whitelabel/registration_page.py
```diff
@@ -1,6 +1,7 @@
 """
 Registration page.
 """
+from edxapp_acceptance.tests.helpers import disable_animations
```
zubair-arbi: Please rearrange imports in alphabetical order.

regression/pages/whitelabel/registration_page.py
```diff
@@ -82,6 +84,13 @@ def register_white_label_user(self, registration_fields, submit=True):
                 }
             )
 
+        # self.wait_for_element_absence(
+        #     '.success',
+        #     'wait for success messages to disappear'
```
zubair-arbi: nit: remove this commented code

regression/tests/helpers/api_clients.py
```diff
@@ -344,56 +345,56 @@ def get_url_from_email(self, matching_string):
...
+        """ Returns an access token and expiration date from the OAuth
+        provider.
         Returns:
-            True or false based on whether user is enrolled
-        """
-        url = '{host}/enrollment/{username},{course_id}'.format(
-            host=self.host,
-            username=username,
-            course_id=course_id
+            (str, datetime)
```
zubair-arbi: (str, datatime): Tuple containing access token string and expiration datetime.

regression/tests/helpers/api_clients.py
```diff
@@ -344,56 +345,56 @@ def get_url_from_email(self, matching_string):
...
+    Base class for creating EdxRestAPiClient instance
     """
+    api_url_root = None
+    append_slash = True
 
-    def __init__(self, host=None, key=None):
-        self.host = host or ENROLLMENT_API_URL
-        self.key = key or ACCESS_TOKEN
+    def __init__(self):
+        assert self.api_url_root
```
zubair-arbi: nit: remove this assertion

regression/tests/helpers/api_clients.py
```diff
@@ -435,6 +436,28 @@ def get_coupon_codes(self, coupon_id):
...
+    """
+    Enrollment API client
+    """
+
+    api_url_root = ENROLLMENT_API_URL
+    append_slash = False
+
+    def is_user_enrolled(self, username, course_run_id):
+        """ Get enrollment details for the given user and course run.
+        Args:
```
zubair-arbi: nit: Please update docstrings format: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html#docstrings

regression/tests/helpers/api_clients.py
```diff
@@ -344,56 +345,56 @@ def get_url_from_email(self, matching_string):
...
+    api_url_root = None
+    append_slash = True
 
-    def __init__(self, host=None, key=None):
-        self.host = host or ENROLLMENT_API_URL
-        self.key = key or ACCESS_TOKEN
+    def __init__(self):
+        assert self.api_url_root
+        access_token, __ = self.get_access_token()
+        self._client = EdxRestApiClient(
```
zubair-arbi: `self._client` -> `self.client`

regression/pages/whitelabel/const.py
kashifch: it is user password for all new and old users, we are using one default value for all passwords

regression/pages/whitelabel/const.py
kashifch: business priority, right now only HarvardMedGlobal tests are active so it makes more sense to have it as default value

regression/tests/helpers/api_clients.py
kashifch: hmm, I copied this code from ecommerce as is, I guess the assertion does make some sense here, we don't want to proceed if url is blank

regression/tests/helpers/api_clients.py
kashifch: hmm, I copied this code from ecommerce as is, I guess underscore is used to differentiate our custom client from any existing clients. Can you elaborate why we need to remove the underscore

regression/pages/whitelabel/const.py
zubair-arbi: then maybe we can do this:
```python
DEFAULT_USER_PASSWORD = os.environ.get('DEFAULT_USER_PASSWORD')
```

regression/tests/helpers/api_clients.py
kashifch: This is a large repo and there would be lots of functions using that format, I will create a seperate pr just for this issue

regression/pages/whitelabel/const.py
zubair-arbi: we can modify it like:
```python
PRIORITIZED_ORG_FOR_TESTING = 'HarvardMedGlobalAcademy'
ORG = os.environ.get('ORG', PRIORITIZED_ORG_FOR_TESTING)
```

kashifch:
@zubair-arbi thanks a lot for reviewing this PR, it is related to white label not enterprise, still it is great to have your input, much obliged :)

regression/tests/helpers/api_clients.py
zubair-arbi: this method is not internal to this base class as in code below we are calling this so we should be using `client` IMHO

regression/tests/helpers/api_clients.py
zubair-arbi: Well we should start the right thing when ever we can :)

zubair-arbi:
@kashifch I thought this work was for enterprise. Thanks for following up on my suggestions.

regression/pages/whitelabel/const.py
kashifch: much better name, will do it

regression/tests/helpers/api_clients.py
kashifch: I guess the underscore is used here because it is called only in derived classes and not available as public import, which is what the "weak internal use indicator" implies or I suppose it implies. Anyway will talk about it in person, would like to learn language's best practices from you as I am just a novice here

regression/pages/whitelabel/const.py
kashifch: on second thought, I will let it be for now, changing the environment variables in different environments could be a tough, some other time may be

regression/tests/helpers/api_clients.py
clintonb: The underscore was there because, in the context from which you copied, `_client` is a private variable.

kashifch: well that settles it, I removed the underscore. I may have misread the code but I thought the _client was being used in inherited classes as well just like I used it

PR #229, opened 2017-11-14
Remove flaky studio textbook test

jzoldak:
This is a flaky test that failed on a pipeline build, and then passed on a subsequent rerun.

Failing build (you need to be logged into jenkins to see):
https://build.testeng.edx.org/job/edx-e2e-tests/851/

Note that you can see the textbook loaded in the screenshot. I suspect that there is a synchronization issue, which most likely derives from the [page object definition](https://github.com/edx/edx-e2e-tests/blob/2585ddc83a4317b30a2142068828ddd16c1c5dfe/regression/pages/lms/lms_textbook.py#L14) for when the page is loaded and ready to go.

@edx/educator-dahlia @dsjen


jzoldak:
jenkins run e2e

PR #234, opened 2017-12-12
WL-1435 Page wait experiment to fix page load failures

kashifch:
Recently we have seen a lot of e2e and microsites test failures with the following message

*BrokenPromise: Promise not satisfied: The document and all sub-resources have finished loading.*

Upon investigating we found out that on stage the page is loaded but somehow the third party requests(there are a lot of requests sent to optimizely, google analytics etc.) get stuck and since the wait for page relies on the document.readyState to become "ready" it throws and error and the test is considered as failure.

I have used an experimental approach where we added an additional check if document does not reach the ready state, instead of throwing error we check if document.readyState="interactive", if so we continue with the test instead of throwing error.

I think this approach can take care of these errors where page is loaded but error is raised due to some third party call getting stuck.

The actual change would have to be done in bok-choy code itself, I have just inherited the class from bok-choy in changed it in my repo so we could also test it's impact.

[WL-1435](https://openedx.atlassian.net/browse/WL-1435)

### DIFF
--- a/regression/tests/helpers/new_page_object.py
+++ b/regression/tests/helpers/new_page_object.py
@@ -1,0 +1,67 @@
+from bok_choy.page_object import PageObject, unguarded
+from bok_choy.promise import Promise, EmptyPromise, BrokenPromise
+
+
+
+class NewPageObject(PageObject):
+
+
+    @unguarded
+    def wait_for_page(self, timeout=30):
+        """
+        Block until the page loads, then returns the page.
+        Useful for ensuring that we navigate successfully to a particular page.
+
+        Keyword Args:
+            timeout (int): The number of seconds to wait for the page before timing out with an exception.
+
+        Raises:
+            BrokenPromise: The timeout is exceeded without the page loading successfully.
+        """
+
+        def _is_document_interactive():
+            """
+            Check the loading state of the document to ensure the document is
+            in interactive mode
+            """
+            return self.browser.execute_script(
+                "return document.readyState=='interactive'")
+
+
+        def _is_document_ready():
+            """
+            Check the loading state of the document to ensure the document and all sub-resources
+            have finished loading (the document load event has been fired.)
+            """
+            return self.browser.execute_script(
+                "return document.readyState=='complete'")
+
+        # Wait for page to laod completely i.e. for document.readyState to
+        # become complete
+
+        try:
+            EmptyPromise(
+                _is_document_ready,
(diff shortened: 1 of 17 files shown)


kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

regression/tests/helpers/new_page_object.py
```diff
@@ -0,0 +1,67 @@
...
+            """
+            Check the loading state of the document to ensure the document and all sub-resources
+            have finished loading (the document load event has been fired.)
+            """
+            return self.browser.execute_script(
+                "return document.readyState=='complete'")
+
+        # Wait for page to laod completely i.e. for document.readyState to
+        # become complete
+
```
kashifch: @jzoldak 

I tried replacing complete state with interactive but it resulted in flakiness as bok-choy tried to perform next actions when the page was not completely ready, so to be on the safe side I tried a round about solution where bok-choy first wait for readyState to become complete but if that doesn't happen in specific time instead of throwing error it checks for interactive state. The tests seems to be passing now but this change would need to go into bok-choy itself as I can override the page object for my customized pages but i won't be able to use it for pages which are installed from platform.  Please take a look and do suggest if there is a better way of achieving that

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

kashifch:
jenkins run microsites

PR #240, opened 2018-01-11
e2e for sandbox

kashifch:
Hi @bill-filler ,

Here is the first draft of changes which enable tests to run on sandbox.

Currently there are some rough edges to fix.

These tests will fail for stage as the org name used in stage is different(MITxPRO) than in sandbox(mitxpro), when the name change is done for both environments these tests should pass.

There is a bug on sandbox due to which course is not shown in dashboard after enrollment, due to this bug majority of test cases would fail

The Environment Variables required to run these tests are updated in readme, you would need to fetch the OAUTH_CLIENT_ID and OAUTH_CLIENT_SECRET from django admin for main ecommerce site.

For TEST_ENV just use helio

Let me know if you run into any issues while running the tests.

One more thing, the bok-choy or selenium is not supported on newer firefox versions so you need to downgrade your firefox to 47.0.1 version or use gecko driver (you would need to put gecko driver in path)

cc @afzaledx @hasnain-naveed

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -165,33 +165,72 @@ To update page objects installed from external repos:
 How to run Whitelabel tests
 ----------------------------
 
-Before running whitelabel tests please ensure that these environment variables are set.
+1. Create a virtual environment for installation/executing tests.
+
+
+2. Change your working directory to `edx-e2e-tests`.
 
 .. code:: bash
 
+    cd edx-e2e-tests/
+
+3. Install the requirements
+
+.. code:: bash
+
+    pip install -r requirements/base.txt
+
+4. Install the page objects for the application from the edx platform repo. This will
+   clone the entire repo into lib/edx-platform so that it can use the page objects and
+   helper methods from common/test/acceptance. We also install capa and xmodule into the
+   virtual environment from common/lib.
+
+
+.. code:: bash
+
+    paver install_pages
+
+5. Set these Environment variables
+
+.. code:: bash
+
+    ==> TEST_ENV (Only required if tests are not running on stage, e.g. for sandbox just assign value `sandbox')
+    ==> TARGET_DNS (Only required if tests are not running on stage, just provide DNS e.g. just provide "helio" for "https://mitxpro-helio.sandbox.edx.org")
+    ==> ORG (Only valid values for now are "MITxPRO" and "HarvardMedGlobalAcademy", if not provided the tests wil run against MITxPRO by default)
     ==> BASIC_AUTH_USER
     ==> BASIC_AUTH_PASSWORD
-    ==> USER_LOGIN_EMAIL
-    ==> USER_LOGIN_PASSWORD
-    ==> COURSE_RUN
-    ==> COURSE_DISPLAY_NAME
-    ==> COURSE_NUMBER
-    ==> COURSE_ORG
     ==> GLOBAL_PASSWORD
-    ==> STAFF_USER_EMAIL
-    ==> TEST_EMAIL_SERVICE
-    ==> TEST_EMAIL_ACCOUNT
-    ==> TEST_EMAIL_PASSWORD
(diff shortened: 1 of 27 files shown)


kashifch:
Just some more info. The real work of making tests generic started some time back due to another issue which reduced the work required for current scenario.

This work is mostly done in prs #217 , #219 , #220 , #227 , #230 , #231 , #232 , #233  

So this time we only have to target the sandbox configuration plus some api changes

There are still some hardcoded parts here, once we have protocols defined for creating ursl, course names and org names these could become cleaner and more generic

bill-filler:
@kashifch 
"There is a bug on sandbox due to which course is not shown in dashboard after enrollment, due to this bug majority of test cases would fail". @afzaledx @hasnain-naveed can you please fix this bug asap otherwise it's impacting the tests

kashifch:
@bill-filler @afzaledx 

I believe these tests are now ready, these are running successfully on stage plus sandbox and not breaking any edx stage e2e tests.

The run on stage is verified here with travis ci, to run these tests locally with sandbox you can just provide the environment vars mentioned in readme file and run the tests.

It would be great if any of you could do a quick review of the code as well

bill-filler:
@kashifch 
I'm trying to run the WL e2e tests against the sandbox but am very confused how to get it setup. The README.rst talks about Devstack configuration. I'm not sure what steps I need to do exactly. Can you add explicit instructions to README.rst that describe exactly the steps needed to run against the Sandbox?

README.rst
```diff
@@ -150,33 +150,70 @@ To update page objects installed from external repos:
...
+.. code:: bash
+
+    cd edx-e2e-tests/
+
+2. Install the requirements
+
+.. code:: bash
+
+    pip install -r requirements/base.txt
+
```
bill-filler: Should add instructions to first create a virtualenv and do the installation from there

kashifch:
done

README.rst
```diff
@@ -150,33 +150,70 @@ To update page objects installed from external repos:
...
-    ==> TEST_EMAIL_ACCOUNT
-    ==> TEST_EMAIL_PASSWORD
-    ==> ACCESS_TOKEN
+    ==> OAUTH_CLIENT_ID (Only required for stage or sandbox is not using the default id)
+    ==> OAUTH_CLIENT_SECRET (Only required for stage or sandbox is not using the default secret)
+    ==> COURSE_RUN (Optional, use only if you are using a value other than default)
+    ==> COURSE_DISPLAY_NAME (Optional, use only if you are using a value other than default)
+    ==> COURSE_NUMBER (Optional, use only if you are using a value other than default)
+    ==> COURSE_ORG (Optional, use only if you are using a value other than specified in ORG var)
+
```
bill-filler: All env variables should have default values. I should only have to specify those that I want to override.
I'm getting errors about STAFF_USER_EMAIL, OAUTH_CLIENT_ID, OAUTH_CLIENT_SECRET not being set.
BASIC_AUTH_* and GLOBAL_PASSWORD should have defaults as well

kashifch: DO you have the latest code, I have already removed the check for STAFF_USER_EMAIL, OAUTH_CLIENT_ID, OAUTH_CLIENT_SECRET, BASIC_AUTH_* .

The only compulsory var right now is GLOBAL_PASSWORD

kashifch: but that was changed due to the change in logic, I would rather have any early warning about a required env rather than tests failing due to an incorrect value. Since the repo is public there is no way we could give a correct value for password in code

kashifch:
jenkins run microsites

kashifch:
Rebase onto master from 2bec7e3c743a11c4b1dd0a7df9cfc31d8835e670

kashifch:
jenkins run microsites

PR #242, opened 2018-02-01
The test_register_and_select_course test is flaky.

feanil:
This test was run on the staging system twice.  The first time it failed.  The second time it succeeded.
Failure: https://build.testeng.edx.org/job/microsites-staging-tests/1284/
Success:  https://build.testeng.edx.org/job/microsites-staging-tests/1285/

### DIFF
--- a/regression/tests/whitelabel/test_otto_enrollment.py
+++ b/regression/tests/whitelabel/test_otto_enrollment.py
@@ -30,17 +30,6 @@ class TestEnrollmentOtto(CourseEnrollmentTest):
         self.course_price = PROF_COURSE_PRICE
         self.total_price = PROF_COURSE_PRICE
 
-    def test_register_and_select_course(self):
-        """
-        Scenario: Otto flow - A registered user is able to register, select a
-        course and make payment for the course using the credit card
-        """
-        self.register_using_api()
-        self.go_to_basket()
-        self.pay_with_cybersource()
-        self.dashboard_page.wait_for_page()
-        self.assert_enrollment_and_logout()
-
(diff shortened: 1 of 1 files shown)


feanil:
jenkins run e2e

kashifch:
The flakiness is not specific to this test, it is the same issue we have
discussed before where a page on staging sometimes does not load fully.It
can happen for any test. A workaround for this is waiting for Jesse's review

I have tried another solution, I have seen this issue locally mostly when I
am using firefox so I created a pr with chrome, lets see if it reduces the
issue

feanil:
@kashifch When the fixes merge, feel free to add the tests back.  In the meantime, this is slowing down the release so I think it should be removed until it is reliable.

kashifch:
sure go ahead and remove the test but again this is not specific to this test, next time it can happen on any other test

kashifch:
here is chrome pr, i will run it 5-6 times and if it pass I think we can try it

https://github.com/edx/edx-e2e-tests/pull/243

if it still fails, I will remove the test

feanil:
@kashifch that sounds good, if we can land that instead, I'm fine with that.  I'll review that PR and we can try to land it today if possible.

kashifch:
@feanil @tuchfarber on second thought just remove the test. I thinK it will try the chrome solution for at least 15-20 times before merging

feanil:
sounds good.  I'll merge once I get a green e2e test build.

kashifch:
So it might extends to tomorrow, meanwhile removing this test could lessen the strain. If chrome test works out I will include it again

PR #244, opened 2018-02-01
Remove flaky studio test.

feanil:
This test failed and then passed on the same version of stage code.

Failed: https://build.testeng.edx.org/job/edx-e2e-tests-pr/562/
Passed: https://build.testeng.edx.org/job/edx-e2e-tests-pr/564/

### DIFF
--- a/regression/tests/studio/test_studio_home.py
+++ b/regression/tests/studio/test_studio_home.py
@@ -7,7 +7,6 @@ from unittest import skipIf
 from bok_choy.web_app_test import WebAppTest
 from edxapp_acceptance.pages.studio.overview import CourseOutlinePage
 
-from regression.pages.lms.lms_courseware import CoursewarePageExtended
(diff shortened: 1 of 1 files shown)


kashifch:
@feanil same issue of document not loading, let's hope we find a solution soon or we end up removing a lot of tests

Wouldn't it be better to skip the problem test instead deleting it

feanil:
@kashifch I'm following the Flaky test process: https://openedx.atlassian.net/wiki/spaces/TE/pages/161427235/Updated+Flaky+Test+Process

I made https://openedx.atlassian.net/browse/EDUCATOR-2250 to resolve fixing this if that team wants to fix it.  So it won't be lost at least.

noraiz-anwar:
Issue with this test is fixed in https://openedx.atlassian.net/browse/WL-1435.
Adding this test back.

PR #251, opened 2018-02-14
Update documentation for stage course upload

michaelyoungstrom:
Right now, the workflow for any developer working in this repo involves running their tests against stage (by default). As this repo becomes more active, this can become problematic, since numerous developers could be manipulating the same course at the same time.

Let's change the workflow to force people to create their own test course. The test course is pretty simple, but since it requires a couple advanced modules (and we already have the tar for the course checked into this repo), the easiest thing to do is set up the course via a course import.

I've updated the tar to change the display name from "Manual Smoke Test Course 1 - Auto" to "E2E Test Course" to be a little more clear about what the course is for.


README.rst
```diff
@@ -64,6 +64,19 @@ variables as necessary. Once done, source the file to get all of the variables i
...
+
+Creating a test course on stage
+-------------------------------
+
+By default, the tests in this repo run against our staging server. If you plan to keep this as the target environment, you must create your own course on stage (this requires an account with staff access). Using your own course will prevent collisions with other developers and ensure that your test results are accurate.
+
+To create the course:
+
+1. Log into studio.edx.org with staff credentials
+2. Create a new course, giving it a unique combination of Organization/ Course Number/ Course Run. The Course Name won't matter here, as it will be overwritten to "E2E Test Course" by the next step.
```
estute: Rather than saying 'a unique combination', perhaps just say use your github username as the org. That way we don't have muliple people using 'foobar' or 'test'. What do you think?

README.rst
```diff
@@ -64,6 +64,19 @@ variables as necessary. Once done, source the file to get all of the variables i
...
+3. Go to the course import page and upload the ``course.tar.gz`` file found in the ``upload_files`` directory of this repo.
+
```
estute: Since they have added unique org/run/number, they will need to reset their environment variables no? We should call this out explicitly.

README.rst
```diff
@@ -64,6 +64,19 @@ variables as necessary. Once done, source the file to get all of the variables i
...
+
```
estute: I think it would also be helpful to describe the test course

michaelyoungstrom:
I have a PR for that here: https://github.com/edx/jenkins-job-dsl/pull/315 @estute

README.rst
michaelyoungstrom: Good point. I think I'll say that for course number instead though. it seems that you must use an existing org.

README.rst
estute: I think so. It would be easy for someone who is new to the repo to overlook the fact that they need to sync these values up.

PR #255, opened 2018-02-16
Fix for LEARNER-4150

raeeschachar:


raeeschachar:
@robrap @feanil @nedo 
This should handle Acquia issue temporarily. Please take a quick look.

regression/tests/studio/test_studio_home.py
```diff
@@ -47,3 +51,56 @@ def test_studio_course_select(self):
...
+        """
+        Verifies that user can click and navigate to studio footer links
+        Terms of Service
+        Privacy Policy
+        """
+        self.studio_home_page.click_terms_of_service()
+        # Temporary condition for Acquia issue
+        self.assertTrue(
+            'Web Site Not Found' in self.browser.title or
+            self.terms_of_service.wait_for_page()
```
robrap: I'm not certain, but I think the test might be more reliable to rely on wait_for_page, and move the `or` check inside `is_browser_on_page` for each page.

raeeschachar:
@robrap done

regression/pages/studio/privacy_policy.py
```diff
@@ -12,6 +12,8 @@ class PrivacyPolicy(PageObject):
     url = LOGIN_BASE_URL + '/edx-privacy-policy'
 
     def is_browser_on_page(self):
-        return "As used in this Privacy Policy" in self.q(
-            css='.field-page-body'
-        ).text[0]
+        # Temporary condition for Acquia issue
```
robrap: This would be a good fix if it is needed.  Two thoughts:
1. I just remembered that yesterday @jaebradley told me that the issue was fixed in Stage, although it persists in other Acquia test environments like Acceptance.  The e2e tests are run against Stage, so we can probably do a simple revert without this change.  @jaebradley: Can you confirm?
2. If I am wrong and the issue is still happening in Stage, please create a clean-up JIRA ticket that is blocked on the Acquia fix and then update these comments to read:
```
# TODO: LEARNER-XXXX: Clean-up temporary condition for Acquia issue
```

jaebradley: @robrap also confirming that all non-prod environments (like `acceptance`) no longer have the intermittent 404 issue

raeeschachar:
@jaebradley @robrap 
Thanks for the update. Will close this PR and revert the other one.

PR #246, opened 2018-02-09
Remove flaky StudioFooterTest.

robrap:
LEARNER-4150

PR #256, opened 2018-02-16
Revert "Remove flaky StudioFooterTest."

raeeschachar:
Reverts edx/edx-e2e-tests#246

### DIFF
--- a/regression/pages/studio/privacy_policy.py
+++ b/regression/pages/studio/privacy_policy.py
@@ -1,0 +1,17 @@
+"""
+Privacy Policy page
+"""
+from bok_choy.page_object import PageObject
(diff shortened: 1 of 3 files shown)


raeeschachar:
@robrap @feanil 
Reverting this. Details [here.](https://github.com/edx/edx-e2e-tests/pull/255)

robrap:
@feanil: Since @mikedikan seemed resistant to having this test removed in the first place, and Mike is out on vacation through next week, I'd propose we do the following:
1. Add this back as-is for now.
2. Have a separate follow-up discussion to better audit what is ok to be a release-blocker and what additional possible process, if any, should be used to ensure everything is properly functioning for those things we decide would not be release blockers.
Are you ok with this?

PR #263, opened 2018-03-09
Use a more complex password when registering new users.

mikix:
This will avoid falling afoul of a site's password complexity requirements.

More complexity can be added if necessary via the NEW_PASSWORD_PREFIX environment variable.


devstack_env
```diff
@@ -10,6 +10,7 @@ CONFIG_ROOT=/edx/app/e2e
 COURSE_NUMBER=E2E-101
 COURSE_RUN=course
 COURSE_DISPLAY_NAME="Manual Smoke Test Course 1 - Auto"
+COURSE_ORG=edX
```
mikix: This change and the USER_LOGIN_EMAIL change below are both unrelated to the main point of this PR. They are just small changes to help when running e2e tests locally -- without these set, "make e2e-tests" in devstack will fail quickly. (They still fail for me with these changes, but due to some browser timeout issue. I can only shave so many yaks in a day.)

PR #273, opened 2018-04-12
TE-2517 Fix test_cancel_upload with existing image

jmbowman:
When you tell selenium to prepare to upload a file, Firefox 59 actually cares if it exists or not; Chrome apparently doesn't until you actually upload it.  I picked an existing image to use, it won't actually be uploaded anyway.


jmbowman:
The only test failing here and the Firefox 59 testing PR is the word cloud one from TE-2515; it seems to be flaky in general, vs. just not working in Firefox 59 (I've seen it pass about 50% of the time with the newer Firefox in devstack).  Tackling that one next.

PR #276, opened 2018-04-13
remove WordCloud e2e test

noraiz-anwar:
https://openedx.atlassian.net/browse/EDUCATOR-2720
Removing this test as it is blocking release to production.


jmbowman:
I should be able to fix and re-add this today, agree removing it is fine for the time being as it's randomly failing for reasons unrelated to what it's supposed to be testing.

noraiz-anwar:
Thanks @jmbowman and @feanil.

PR #280, opened 2018-04-18
Fix travis by pinning docker-ce version

michaelyoungstrom:
docker ce version 18.04.0 was prompting us before logging in, which resulted in a timeout and a failed build. Someone has reported the issue here: https://github.com/docker/cli/issues/1002 . For now, let's pin the version to 18.03.0.


thaJeztah:
The interactive prompt / confirmation was reverted in https://github.com/docker/cli/pull/1008, which is part of Docker 18.05 and up.

If you want to stick to the 18.03 version; I would recommend updating this to 18.03.1, which is a patch release for 18.03 and contains various bugfixes

PR #281, opened 2018-04-30
Update privacy language

efischer19:
Tests are failing on this page: https://build.testeng.edx.org/job/edx-e2e-tests/1244/

Due to this change: https://github.com/edx/edx-mktg/pull/2163

If I'm not mistaken, this text should be valid for both versions of the page.

FYI @AlasdairSwan @dianakhuang


efischer19:
Updated to a slightly shorter string to satisfy quality checkers

efischer19:
ugh, looks like my force-push-to-fix-quality resulted in a second build that had to wait for the first one to complete before it started

ETA of ~35 minutes before jenkins/e2e passes

PR #282, opened 2018-04-30
Wait for button to appear

efischer19:
Should prevent flaky test failures like [this one](https://build.testeng.edx.org/job/edx-e2e-tests/1247/testReport/junit/regression.tests.studio.test_course_import/TestCourseImport/test_course_updated/)


efischer19:
The first commit addressed build number 1247, the second addresses a similar issue in 1248.

PR #283, opened 2018-05-22
bump bok-choy version to 0.7.3

kashifch:
bump bok-choy version to 0.7.3

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -3,7 +3,7 @@ path.py==7.2
 requests==2.0.1
 nose==1.3.7
 Paver==1.2.4
-bok-choy==0.7.2
+bok-choy==0.7.3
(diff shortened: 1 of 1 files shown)


kashifch:
jenkins run microsites

kashifch:
jenkins run e2e

feanil:
jenkins run all

estute:
jenkins run e2e

estute:
@kashifch this pr has removed the .gitkeep file in the logs directory. Without it, the directory is empty and therefore git no longer indexes it. For the short term, please add the empty file back in. Since the directory only seems to store artifacts created at run-time, I think we should create the directory itself at run-time (but we can do that in a separate PR)

kashifch:
@feanil @estute thanks for the help

PR #290, opened 2018-06-27
remove hms

kashifch:


feanil:
@kashifch  I think we can merge this without waiting for the tests because they'll have to be run as a part of the pipeline anyway.

PR #299, opened 2018-07-26
remove tests for edit start date button

MichaelRoytman:


MichaelRoytman:
These will be replaced in the following ticket as bok choy tests: https://openedx.atlassian.net/browse/EDUCATOR-3255.

PR #364, opened 2018-12-23
Fix and un-skip Studio login/logout tests

nasthagiri:


robrap:
@nasthagiri: I dropped myself as a reviewer since Diana already completed it, unless you think this requires two reviews.

PR #366, opened 2019-01-03
Skip test_studio_course_select because it is timing out

cpennington:


cpennington:
jenkins run microsites

cpennington:
jenkins run microsites

cpennington:
This was probably failing because the courses it was looking for ended on 12/31/2018. I updated them to end on 12/31/2030.

PR #367, opened 2019-01-03
Don't make the WL_E2E default course run change yearly

cpennington:


cpennington:
Merged as part of https://github.com/edx/edx-e2e-tests/pull/369

PR #375, opened 2019-03-07
Enteprise recovery email feature tests

zrana:
Added tests for enterprise recovery email feature.

### DIFF
--- a/regression/pages/enterprise/user_account.py
+++ b/regression/pages/enterprise/user_account.py
@@ -1,6 +1,7 @@
 """
 User account page
 """
+from selenium.webdriver.common.keys import Keys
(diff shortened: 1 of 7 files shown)


regression/pages/enterprise/user_account.py
```diff
@@ -40,3 +41,24 @@ def unlink_idp_account(self, idp_css_id):
...
+            email
+        """
+        self.wait_for_element_visibility('#field-input-secondary_email', 'Secondary Email field is shown')
+        elem = self.q(css='#field-input-secondary_email').results[0]
+        elem.clear()
+        elem.send_keys(email)
+        elem.send_keys(Keys.ENTER)
+
+    @property
+    def get_user_email(self):
```
kashifch: when using property you don't need to add get with method name

PR #376, opened 2019-03-12
add e2e test for submitting a graded problem

MatthewPiatetsky:


regression/tests/lms/test_graded_problem.py
```diff
@@ -0,0 +1,46 @@
+"""
+End to end tests for Instructor Dashboard.
```
MatthewPiatetsky: note to self to fix comment

MatthewPiatetsky:
the method has a default timeout

PR #379, opened 2019-03-19
Undo "Revert of "add e2e test for submitting a graded problem""

MatthewPiatetsky:
Undo Revert edx/edx-e2e-tests#378 once cause of this failure https://build.testeng.edx.org/job/edx-e2e-tests/2140/testReport/junit/regression.tests.lms.test_graded_problem/GradedProblemTest/test_graded_problem/ is identified

Cause was wrong course being accessed from the dashboard, fix was accessing the course page directly. The error was only present in the edx-e2e-tests job, not the edx-e2e-tests-pr job, because the test worked for the E2E-1001 course but not the E2E-1000 course.

Reproduced error using the E2E-1000 course with this build https://build.testeng.edx.org/job/edx-e2e-tests-pr/847/parameters/  
Successful build with fixed code and same parameters  
https://build.testeng.edx.org/job/edx-e2e-tests-pr/849/parameters/


MatthewPiatetsky:
jenkins run all

PR #389, opened 2019-11-13
When creating a new unit, the display name editor should be defocused before moving on to other actions on the page

davestgermain:
This is needed after the changes in https://github.com/edx/edx-platform/pull/21961

### DIFF
--- a/regression/tests/common/test_html_component.py
+++ b/regression/tests/common/test_html_component.py
@@ -62,6 +62,7 @@ class StudioLmsAdvancedComponentTest(StudioLmsComponentBaseTest):
         """
         Verifies that user can add Word Cloud component on Studio and LMS
         """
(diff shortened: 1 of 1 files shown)


regression/tests/common/test_html_component.py
```diff
@@ -92,7 +95,7 @@ def test_word_cloud_advanced_component(self):
 
         # View Live
         self.unit_container_page.view_live_version()
-        self.lms_courseware.wait_for_page()
```
jmbowman: I get the change above, but why did this need to be removed?

davestgermain: I don't know, but I saw that the page was hanging on the wait, even though everything was loaded. I assume it's timing related?

PR #390, opened 2019-11-13
Speed up course selection on studio home page.

davestgermain:
This trims 14 minutes off of the total test time.

### DIFF
--- a/regression/pages/studio/studio_home.py
+++ b/regression/pages/studio/studio_home.py
@@ -31,15 +31,11 @@ class DashboardPageExtended(DashboardPage):
         """
         Selects the course we want to perform tests on
         """
-        course_title_selector = '.course-link h3'
-        self.wait_for_element_presence(course_title_selector,
-                                       'Course link presence')
-        course_names = self.q(css=course_title_selector)
-        for vals in course_names:
-            if course_title in vals.text:
-                vals.click()
-                return
-        raise BrokenPromise('Course title not found')
+        query = self.q(xpath='//h3[contains(text(), "{}")]'.format(course_title)).first
+        if course_title in query.text:
+            query.click()
+        else:
+            raise BrokenPromise('Course title not found')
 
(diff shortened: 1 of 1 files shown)


regression/pages/studio/studio_home.py
```diff
@@ -31,14 +31,10 @@ def select_course(self, course_title):
...
         """
-        course_title_selector = '.course-link h3'
-        self.wait_for_element_presence(course_title_selector,
-                                       'Course link presence')
-        course_names = self.q(css=course_title_selector)
-        for vals in course_names:
-            if course_title in vals.text:
-                vals.click()
-                return
+        elements = self.browser.find_elements_by_xpath('//h3[contains(text(), "%s")]' % course_title)
```
cpennington: You could use `self.q(xpath='//h3[contains(text(), "{}")]'.format(course_title)).first.click()` to get promises and retries built in so that you don't get stale element exceptions.

regression/pages/studio/studio_home.py
```diff
@@ -31,15 +31,11 @@ def select_course(self, course_title):
...
-        self.wait_for_element_presence(course_title_selector,
-                                       'Course link presence')
-        course_names = self.q(css=course_title_selector)
-        for vals in course_names:
-            if course_title in vals.text:
-                vals.click()
-                return
-        raise BrokenPromise('Course title not found')
+        query = self.q(xpath='//h3[contains(text(), "{}")]'.format(course_title)).first
+        if course_title in query.text:
```
cpennington: Why wouldn't this find the course title, given that that's what your Xpath searched for?

davestgermain: It would, but there needs to be some way to verify that the element was found, otherwise the BrokenPromise will never be raised. The `click()` method doesn’t return anything, so a one-liner wouldn’t be enough.

PR #392, opened 2019-11-28
Fix for too many redirect issue

ziafazal:
Fix for `Firefox has detected that the server is redirecting the request for this address in a way that will never complete.`

### DIFF
--- a/regression/tests/common/test_html_component.py
+++ b/regression/tests/common/test_html_component.py
@@ -12,7 +12,7 @@ from regression.pages.lms.utils import get_course_key
 from regression.pages.studio.course_outline_page import CourseOutlinePageExtended
 from regression.pages.studio.studio_home import DashboardPageExtended
 from regression.pages.studio.unit_page import UnitPageExtended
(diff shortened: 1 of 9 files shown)


crice100:
@ziafazal Do you happen to be know why this is necessary? Did something change so that we require it now?

ziafazal:
@crice100 I think either of these could have been changed

1. e2e test user `edx-e2e-tests@edx.org` is no more a global staff
2. domain of the session cookie on LMS side return by `user_api/v1/account/login_session` API.

crice100:
Thanks Zia. Since we don't seem to know exactly why this change would be required and we've already merged [the revert PR](https://github.com/edx/edx-platform/pull/22413/), should we wait to see if that revert fixes the test before merging this?

ziafazal:
@crice100 yeah I'm open to keep my changes away from stage to see if e2e tests pass without them.

estute:
jenkins run e2e

cpennington:
jenkins run e2e

PR #403, opened 2020-04-16
fix logout is_browser_on_page

robrap:
Need to return False, rather than an exception, so wait_for_page has
a change to actually wait.

### DIFF
--- a/regression/pages/whitelabel/logout_page.py
+++ b/regression/pages/whitelabel/logout_page.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import
 
 import os
 from bok_choy.page_object import PageObject
+from selenium.common.exceptions import WebDriverException
 
 from regression.pages.whitelabel import ECOM_URL_WITH_AUTH
 from regression.pages.whitelabel.home_page import HomePage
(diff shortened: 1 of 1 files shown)


regression/pages/whitelabel/logout_page.py
```diff
@@ -27,5 +28,9 @@ def is_browser_on_page(self):
         """
         home_page = HomePage(self.browser)
 
-        return ("you have signed out" in self.browser.page_source.lower()) or \
-            home_page.is_browser_on_page()
+        try:
+            return ("you have signed out" in self.browser.page_source.lower()) or \
+                home_page.is_browser_on_page()
+        except WebDriverException:
```
robrap, line 34: **Reviewer Notes**:
* First tried `TypeError`, but that failed with `WebDriverException`.
* Next tried `Exception`, which worked, but failed quality.
* Now trying `WebDriverException`.
If this last one doesn't pass microsites e2e, worst case, use `Exception` and disable quality error.

robrap: `WebDriverException` seemed to do the trick, so I am resolving this.

PR #446, opened 2021-02-19
fix: Delete flaky test DataDownloadTest::test_generate_grade_report

timmc-edx:
Deleted according to flaky test process:
https://openedx.atlassian.net/wiki/spaces/TE/pages/161427235/Flaky+Test+Process

Flaky test ticket: CR-3320


feanil (commented):
Looking at the stack trace, it looks like the test actually failed while trying to login to the LMS as a part of the setup and timing out, which seems not related to this test.  It seems like it is a general problem of this test suite.

If we think it's specific to that setup though, then we should probably delete both of these tests or we should fix the setup.

timmc-edx:
Hmm, you're right. It's not the test itself that's failing.

I don't see anything in the setup that looks dodgy. But this specific test has failed a few times. Maybe there's something about test order and needing to let the browser warm up?

MatthewPiatetsky:
@feanil @timmc-edx I saw a failure of this test on this job https://build.testeng.edx.org/job/edx-e2e-tests/5886/testReport/junit/regression.tests.lms.test_instructor_dashboard/DataDownloadTest/test_generate_problem_grade_report/

but I reran and the test passed
does this mean this test is flaky after all?

timmc-edx:
Hmm, that's the other test in the same test class.

matthugs:
I have another stack trace to throw into the mix. This is a very prolifically failing test.
```Error Message
bok_choy.promise.BrokenPromise: Promise not satisfied: Waiting for downloadable report
Stacktrace
self = <regression.tests.lms.test_instructor_dashboard.DataDownloadTest testMethod=test_generate_grade_report>

    def test_generate_grade_report(self):
        """
        Tests the Generate Grade Report task
        """
        initial_reports = self.data_download_section.report_download_links.text
        initial_pending_tasks = self.data_download_section.get_number_of_pending_tasks
        self.data_download_section.generate_grade_report_button.click()
>       self.data_download_section.wait_for_available_report()

regression/tests/lms/test_instructor_dashboard.py:76: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
venv/lib/python3.8/site-packages/bok_choy/page_object.py:120: in wrapper
    return method(self, *args, **kwargs)
edxapp_acceptance/pages/lms/instructor_dashboard.py:1146: in wait_for_available_report
    EmptyPromise(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <bok_choy.promise.EmptyPromise object at 0x7f78cf3fb610>

    def fulfill(self):
        """
        Evaluate the promise and return the result.
    
        Returns:
             The result of the `Promise` (second return value from the `check_func`)
    
        Raises:
            BrokenPromise: the `Promise` was not satisfied within the time or attempt limits.
        """
        is_fulfilled, result = self._check_fulfilled()
    
        if is_fulfilled:
            return result
>       raise BrokenPromise(self)
E       bok_choy.promise.BrokenPromise: Promise not satisfied: Waiting for downloadable report

venv/lib/python3.8/site-packages/bok_choy/promise.py:105: BrokenPromise
```
from https://build.testeng.edx.org/view/e2e-tests/job/edx-e2e-tests/5909/testReport/regression.tests.lms.test_instructor_dashboard/DataDownloadTest/test_generate_grade_report/

timmc-edx:
OK yeah, and that one isn't failing in setup. I was also wondering if there was a problem here with browser initialization being slow, but this isn't even the first test to run. I'm going to change the PR to remove the whole test class.

crice100 (approved):
I'd suggest opening a ticket to track this deletion (per the process), but I fully support this deletion as it just failed again.

https://build.testeng.edx.org/job/edx-e2e-tests/5917/

timmc-edx:
Ticket already filed: CR-3320

I'll merge tomorrow morning.

feanil (approved):

PR #449, opened 2021-03-05
Update to Python3 after AMI Upgrade

mzulqarnain1:
e2e tests are failing on stage after AMI upgrade, @syedimranhassan thinks this change can fix that problem.


syedimranhassan (approved):
LGTM

adzuci:
We needed to rollback to 16.04 workers so I reverted this just in case it caused issues: https://github.com/edx/edx-e2e-tests/pull/453

PR #459, opened 2021-03-23
test: Two tests are failing on ubuntu20. Skipping both temporarily.

awais786:
https://openedx.atlassian.net/browse/BOM-2460

skip e2e tests both are failing on Ubuntu 20.04

### DIFF
--- a/regression/tests/common/test_html_component.py
+++ b/regression/tests/common/test_html_component.py
@@ -1,14 +1,15 @@
 """
 End to end tests for HTML Components
 """
-
+from unittest import skip
 from uuid import uuid4
 
 from bok_choy.web_app_test import WebAppTest
 
 from regression.pages.lms.lms_courseware import CoursewarePageExtended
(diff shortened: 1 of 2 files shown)


nadeemshahzad (approved):

regression/tests/common/test_html_component.py
```diff
@@ -1,14 +1,15 @@
...
+from unittest import skip
 from uuid import uuid4
 
 from bok_choy.web_app_test import WebAppTest
 
 from regression.pages.lms.lms_courseware import CoursewarePageExtended
 from regression.pages.lms.utils import get_course_key
-from regression.pages.studio.course_outline_page import CourseOutlinePageExtended
+from regression.pages.studio.course_outline_page import \
+    CourseOutlinePageExtended
```
jmbowman, line 12: Doesn't seem like this line break should be needed.

jmbowman (approved):

PR #464, opened 2021-03-25
test: Unit was not loading be default. Manualy load the sub-section a…

awais786:
This test is failing on ubuntu20 jenkin worker

Test Case steps:

- Login form LMS.
- Click `View in Studio` button.
- In Studio Add Section, sub-section, unit ( word cloud ), publish it.
- Click on `View Live Version` (Unit appears in lms successfully). 
- Now Click `View in Studio` button. ( In test case unit fails to load. Following screenshot shows sub-section is not collapsed by default)
- Previous failed build [results](https://build.testeng.edx.org/view/e2e-tests/job/edx-e2e-tests/6142/).

Screen Shot 2021-03-25 at 6 11 39 PM

But doing this manually on stage unit loads successfully. 

Proposed Fix.
I trigger the click on sub-section and unit.

Note: Issue is appearing on stage only. Locally it is working fine.
https://openedx.atlassian.net/browse/BOM-2460

### DIFF
--- a/regression/tests/common/test_html_component.py
+++ b/regression/tests/common/test_html_component.py
@@ -1,7 +1,6 @@
 """
 End to end tests for HTML Components
 """
-from unittest import skip
 from uuid import uuid4
 
 from bok_choy.web_app_test import WebAppTest
@@ -54,7 +53,6 @@ class StudioViewTest(StudioLmsComponentBaseTest):
     HTML Components tests related to 'studio view' of component.
     """
 
-    @skip("BOM-2460: Test is failing on Ubuntu 20.04.")
     def test_unit_studio_view(self):
         """
         Scenario: To test studio view of component from LMS
@@ -79,6 +77,11 @@ class StudioViewTest(StudioLmsComponentBaseTest):
         self.lms_courseware.go_to_section(section_name, subsection_name)
         # View unit in the studio
         self.lms_courseware.view_unit_in_studio()
+        # View the course in studio.
+        self.studio_course_outline.visit()
+        self.studio_course_outline.click_sub_section()
+        self.studio_course_outline.click_unit_button()
+
         self.unit_container_page.wait_for_page()
         # Correct unit component should open.
         self.assertEqual(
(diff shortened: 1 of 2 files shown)


awais786:
jenkins run all

regression/tests/common/test_html_component.py
```diff
@@ -79,6 +77,11 @@ def test_unit_studio_view(self):
         self.lms_courseware.go_to_section(section_name, subsection_name)
         # View unit in the studio
         self.lms_courseware.view_unit_in_studio()
+        # View the course in studio.
+        self.studio_course_outline.visit()
```
jmbowman, line 81: Do you need the `visit()` call?  I thought the test was ending up on the right page but just not expanding the appropriate subsection and unit; `visit()` takes the browser to this page even if it's currently on a different page entirely.

jmbowman (commented):
Thanks for diagnosing the problem!  It's a little concerning that the behavior seems to change when we switch to running the test on Ubuntu 20.04, but I guess I'm ok with this fix as long as we're not actually ending up on completely the wrong page.

Have you tried again since stage was also upgraded to 20.04, to see if the test still hits the problem without this fix?

regression/tests/common/test_html_component.py
awais786: `self.lms_courseware.view_unit_in_studio()` loads this page and you can see sections are appearing as bread-crumbs and its not possible to click them.

Screen Shot 2021-03-25 at 8 05 36 PM

------------------------------------------------------------------------------------------------------------
but `self.studio_course_outline.visit()` loads same page in collapsable form.


Screen Shot 2021-03-25 at 8 07 14 PM

awais786:
> Thanks for diagnosing the problem! It's a little concerning that the behavior seems to change when we switch to running the test on Ubuntu 20.04, but I guess I'm ok with this fix as long as we're not actually ending up on completely the wrong page.
> 
> Have you tried again since stage was also upgraded to 20.04, to see if the test still hits the problem without this fix?

https://github.com/edx/edx-e2e-tests/pull/465 I have made the PR and both tests failing.

awais786:
jenkins run all

awais786:
@jmbowman Please review again.

regression/tests/common/test_html_component.py
awais786: This is verifying the locators.
```
self.assertEqual(
            get_data_locator(self.unit_container_page),
            data_locator, 'Correct component is opened'
        )
```

iamsobanjaved (approved):

PR #510, opened 2021-11-17
fix: studio privacy policy test fix

robrap:
The privacy policy text changed, and this fixes to
a different part of the page that hopefully won't
change often.


adzuci (approved):
:shipit:

PR #517, opened 2021-12-20
chore: Remove SETUPTOOLS_USE_DISTUTILS="stdlib"

adzuci:
This was added as a temporary measure to work around a setuptools 50.0.0 issue.  We may need to add it back for 60.0.0, but I'd like to see if it can be removed currently.

Note: I am not confident in the impacts of this change.


matthugs (approved):
clean all the things

